PR #7, opened 2026-09-03
Prepare the Postgres codebase for a second database engine

Azumi0:
## Description

This is **stage 1 of two**: preparation for running Filadex as a single container with no Postgres server, by supporting SQLite alongside it. **Stage 1 contains no SQLite**, and that is deliberate — every item here has to stand on its own merit against the Postgres-only codebase. You should be able to accept all of it without agreeing that SQLite is a good idea, or that it will ever happen. If it is abandoned, nothing here has to be reverted.

The reasoning, including what was decided and what was deliberately left alone, is recorded in `docs/adr/0001-prepare-the-postgres-codebase-for-a-second-engine.md`. That ADR is the best entry point for reviewing this.

**What it does, in short:**

- **A test suite where there was none** — 243 tests against a real Postgres, no in-memory fake. `TEST_DATABASE_URL` selects the server, or a throwaway `postgres:15-alpine` container is started for the run.
- **All database access goes through `IStorage`.** Within the application, `server/db.ts` is now imported by `server/storage.ts` and nothing else; `MemStorage`, which was unused and untested, is gone.
- **`shared/schema.ts` now describes the database that actually exists.** Reconciling it with what a real deployment contains produced 57 statements of difference. Most consequentially, ten timestamp columns would have been narrowed from `timestamptz`, discarding each value's UTC offset.
- **Generated Drizzle migrations, with a baseline for existing installs.** `docker-entrypoint.sh` went from 14 `psql` invocations, three inline `ALTER TABLE` checks and eleven scripts re-run on every start, to one call to `scripts/migrate.ts`. Existing databases run the legacy chain once and are then baselined; the legacy scripts are kept forever, because deleting them means an older installation can never upgrade, only start over.
- **`npm run db:verify-upgrade`**, run in CI, proves the upgrade rather than asserting it: it builds a pre-Drizzle install from frozen DDL, upgrades it, and checks that every row survives byte-identical and that the result matches a fresh install.

**Eight bugs were found and fixed along the way**, each pinned by a test. They are listed in the ADR with *how* they were found, which is the actual argument for this work. One is worth your attention independently of this PR: **the email migration verified every pending registration on every container restart**, letting anyone bypass email verification by waiting for a restart. That one is live in the released version and is not introduced here.

**Deliberately not done**, and recorded in `TODO.md` instead, because each is a product decision rather than a refactor: filaments duplicating catalog names as free text, the two divergent sharing endpoints, the 8-versus-6 password split, and the dead `filaments.created_at`/`updated_at` columns.

**One documented deviation from CONTRIBUTING.md:** the files under `migrations/pg/` do not end with a newline. They cannot. `scripts/migrate.ts` baselines an existing installation by writing the hash of the migration file, taken over its bytes, exactly as drizzle's own migrator does — a trailing newline changes that hash, and every already-baselined deployment would then see a migration it does not recognise. The reasoning is written next to the baseline table in the ADR so nobody "fixes" it later.

## Type of change

- [x] Documentation update
- [x] Code refactoring

## How Has This Been Tested?

Test configuration: Node with Docker available; the suite starts its own `postgres:15-alpine` unless `TEST_DATABASE_URL` points at an existing server.

- [x] `npx vitest run` — 9 files, 243 tests, all passing
- [x] `npm run check` — clean; `tsconfig.json` now covers `scripts/**` and `migrations/**` as well as the application
- [x] `npm run db:verify-upgrade` — all six checks pass: every row survives byte-identical, an upgraded database matches a fresh install, both match `shared/schema.ts` per `drizzle-kit`, and re-running changes neither a row nor the schema
- [x] `npm run build` — succeeds
- [x] Migration run manually against an empty database, then run a second time, confirming the schema is created, the `email_settings` row exists, and the second run is a no-op
- [x] CI runs the test suite and the upgrade verification on every pull request

## Checklist:

- [x] My code follows the style guidelines of this project — with the one documented exception above
- [x] Any dependent changes have been merged and published in downstream modules — none

https://claude.ai/code/session_01Js9C1arB8X188n7jCffhHs


the-luap:
## Review of this branch, and a commit pushed onto it

I reviewed this PR and pushed the fixes as `4bdcfa8` directly onto `stage1-cleanup` (the PR allows maintainer edits). Apologies for touching your branch rather than opening a follow-up — say the word and I'll move it out into its own PR instead.

**The headline: the claim in the description holds.** I checked `shared/schema.ts` column-by-column against the legacy DDL, and it's accurate — constraint names match Postgres's implicit naming, the plain-vs-`timestamptz` split matches which script created each column. The email-verification fix is correct and complete: `introducingEmailVerification` gates on the column being absent, which is right for all three paths, and it's genuinely pinned by the seeded `carol` row. `MemStorage` removal is complete, `server/db` is imported only by `storage.ts`, and the Dockerfile doesn't reintroduce GH#5.

The findings are about what happens when the legacy chain *doesn't* do what `baseline()` then asserts it did.

### The one that mattered

`baseline()` ran unconditionally after a chain with three paths that report success while doing nothing — `add_user_id_column.ts:78` catches the `ALTER TABLE` error and continues before logging "completed successfully"; `drop_filament_type_columns.ts:37,58` `return` rather than throw when rows are unlinked or the `ALTER` hits `42501`. That's the ownership shape `992d92f` and `edf0712` were written for.

Because baselining is what stops the chain from ever running again, such a database is **unrecoverable by restart** — where the old entrypoint re-ran the chain each boot and self-healed once ownership was fixed.

`assertMatchesBaseline` now parses 0000's own SQL (15 tables, 128 columns, types) and refuses to record the baseline if the live schema differs, leaving the database untouched so a retry works. It's parsed rather than listed, so it can't go stale. The same guard catches two other cases: base tables an old install lacks (the chain creates none of the eight — `user_sharing` is the concrete instance), and the plain `timestamp` that `drizzle-kit push` leaves where 0000 has `timestamptz`.

I verified it fires, with a throwaway harness: it refuses a database with `user_sharing` dropped and one with `colors.created_at` narrowed, writes no journal row in either case, and still baselines a healthy database.

### The rest

- **`hasDrizzleJournal` tested the table, not its rows.** `baseline()`'s `CREATE TABLE` and `INSERT` autocommit separately; dying between them meant every later start ran 0000 against a populated database and failed identically forever. Now gated on a row count.
- **No advisory lock.** Two containers starting together both acted. Now `pg_advisory_lock`, held on one client for the whole run.
- **`PUT /api/users/:id` validated nothing** while `POST` gained `usernameSchema` — so a username the system refuses to create could still be reached by creating a valid one and renaming into it. It now uses `adminUpdateUserSchema`. The admin form had only its *password* rule updated, so `max.mustermann` passed client validation and came back as an untranslated toast; it now mirrors `usernameSchema` in full, reusing the existing `auth.*` keys. Both endpoints still accept the `"true"`/`1` booleans they accepted while unvalidated, and blank still means "leave alone".
- **`upsertUserSharing` updated an arbitrary duplicate** (`select()` with no `ORDER BY`/`LIMIT`). Older releases left a duplicate global row on each toggle and `getPublicUserSharing` takes any row with `is_public`, so `POST /api/sharing` couldn't switch off a collection that `setUserSharing` had already been fixed to handle. It now collapses duplicates, oldest wins.
- **`verify-upgrade`'s row check didn't cover the backfill.** Its window opens *after* the chain has run — correct for what it tests, since the old entrypoint ran the chain every boot, but it puts `add_filament_types` and `drop_filament_type_columns` outside the measurement entirely. The seed can't close it either: it inserts through `shared/schema.ts`, which can't express a flat pre-migration `filaments` row. Added `verifyBackfill`, which builds its own database, stops before the chain, seeds flat rows through raw SQL, and compares each spool's product identity across the transformation — including that two spools of one product share a type and the same product owned by two users doesn't.
- Also in `verify-upgrade`: the table list is derived from the schema instead of hardcoded (a future table would have been silently excluded while still printing PASS), and a failure throws instead of `process.exit`, which was skipping the `finally` and leaking both testcontainers.

### Two behaviour changes that need release notes

Both are consequences of fixes here, and both reach users on upgrade with no action on their part. Written up in the ADR:

1. **Case-insensitive material matching can make a spool public that wasn't.** A user sharing catalog `PLA` who owns a spool whose free-text material reads `pla` was not sharing it before and is now — on an endpoint that needs no authentication.
2. **Timestamps read back correctly, so they read back differently.** On a non-UTC Postgres session, `createdAt`/`lastLogin` shift by the offset across the upgrade. The new value is the right one. Nothing tests it, because the containers are UTC.

### One doc correction

The ADR justified the no-trailing-newline rule by saying the hash is "what makes the baseline recognisable to later `drizzle-kit` runs". In drizzle-orm 0.39.1 the migrator selects `order by created_at desc limit 1` and compares `created_at` — the `hash` column is written and never read. The convention is still right (it's undocumented internal behaviour of a dependency, and matching drizzle's own writer keeps the row honest), but the stated invariant doesn't exist and would mislead whoever reasons from it. Rewritten to say why it actually holds.

### One open question for you

`usernameSchema` rejects non-ASCII, so `müller` cannot be registered at all. I kept that rule and only made the client agree with it, since loosening it is a product decision rather than a review fix — but given the German UI, is that intended? If not, the regex is the thing to revisit and I'm happy to follow up.

### Verification

- `npx vitest run` — 9 files, **249 tests** passing (your 243 plus 6 new: the `PUT` validation cases and the duplicate-sharing collapse)
- `npm run check` — clean
- `npm run db:verify-upgrade` — all **7** checks pass, including the new backfill one

Azumi0:
Pushing onto the branch was the right call — it's on the PR head now as `020ec0a`, rebased onto our tip. No need to move it out.

I checked the findings rather than taking them as read. They hold, including the ones I'd have been least happy to be wrong about:

- **The baseline parser.** Ran `baselineShape` standalone against `0000_right_mathemanic.sql`: 15 tables, 128 columns, and — the part I actually wanted to know — **zero** columns fall through the `DATA_TYPES` map. An unmapped type would be skipped silently and quietly narrow the guard, so it's worth a regression test if a future migration introduces `varchar` or an enum.
- **`assertMatchesBaseline` runs before `CREATE SCHEMA drizzle`**, so a refusal genuinely leaves the database untouched. That's the property the whole fix rests on.
- **The drizzle correction is right, and I'd have believed the original.** `pg-core/dialect.cjs` selects `id, hash, created_at ... order by created_at desc limit 1` and gates on `Number(lastDbMigration.created_at) < migration.folderMillis`. `hash` is selected and never compared.
- **Advisory lock**: the pool is constructed with no `max`, so the default of 10 applies and holding one client for the run can't starve `main()`.

`npx vitest run`, `npm run check` and `npm run db:verify-upgrade` reproduce the numbers as stated.

### On the username question — the premise needs one correction

You wrote that you kept the rule. The regex is older than this PR (`5ce56e8`, untouched here), but it only ever bound **self-registration**. On `main`, `POST /api/users` and `PUT /api/users/:id` both destructure `req.body` with no validation at all, and the admin form asked only for `min(3)`. So an admin *could* create `müller` on `main`, and after this PR they cannot. That's an upgrade-visible change, and it wasn't in the release-notes section — it is now.

### And it had a consequence neither of us caught

`user-management-modal.tsx:192` prefills `username: user.username` on edit. For an install already holding a `müller` account, an admin opening it to reset the password or toggle the admin flag resubmits the held name — which now fails validation, client-side *and* server-side, since `omitIfBlank` only skips empty strings. That account becomes unadministrable. Login was never at risk (`/api/auth/login` looks the user up without validating), so it's a lockout from administration, not from access.

Fixed in `82200e5`: **the rules apply to a name being set, not to one already held.** The endpoint applies `usernameSchema` only when the submitted name differs from the current one; the form exempts exactly that name. Everything the tightening was for survives — renaming into a refused name is still refused, for this user and any other, and recasing counts as setting a new name so it's checked too. The only exit for such an account is a rename to an acceptable name. Four tests, including the two that pin what must *stay* refused.

### The product decision I left alone

I didn't touch the regex. Widening it is a product call that's yours, and I'd rather not make it inside a migration-cutover PR. One thing worth weighing if you do: usernames are compared with `LOWER()` (`server/db/predicates.ts`), and its behaviour on non-ASCII depends on the database's collation — so uniqueness and login would want checking together, which is exactly the pair that disagreed last time. Recorded in `TODO.md` with that note.

Verification after the fix: `npx vitest run` — 9 files, **253 tests** (your 249 plus 4); `npm run check` clean; `npm run db:verify-upgrade` — all 7 pass.

https://claude.ai/code/session_01Js9C1arB8X188n7jCffhHs

the-luap:
Checked `82200e5` against the code. The bug is real and the fix is the right shape — thank you for catching it.

I'd tightened `PUT` to close a hole and didn't follow the consequence into the edit form. Your framing is the correct one: **the rules apply to a name being set, not to one already held.** Exempting exactly the held name keeps every case the tightening existed for — renaming into a refused name is still refused, for this user and any other, and recasing counts as setting a new name so it's still checked. The `superRefine` on the client mirrors the endpoint's condition rather than approximating it, which is what keeps the two from drifting again.

Your correction to my premise is also right, and it's the part I'd have kept getting wrong: I wrote "I kept that rule" when the rule only ever bound self-registration. Admins really could create `müller` on `main`, and after this PR they can't. That belonged in the release-notes section and I'd missed it.

## On the username regex — I opened #8, then took it back to draft

I went ahead with the widening as a separate PR against `main` (it doesn't depend on this one; `usernameSchema` predates it). Two things came out of it worth reporting back here.

**The character class is Latin script, not `\p{L}`.** `\p{L}` also admits Cyrillic `а` (U+0430), which renders identically to Latin `a` — so `аdmin` and `admin` would be two accounts indistinguishable wherever a username is shown, including the public sharing page. Nothing detects confusables and case-insensitive uniqueness can't help, since the strings differ in every comparison the database can make. Latin script covers every umlaut, eszett and accent the German and English UIs can produce without opening that.

Widening to diacritics also makes NFC normalisation part of the rule rather than an afterthought: `ü` is either U+00FC or `u` + U+0308 depending on the keyboard, and left alone those are different strings — two accounts that display the same name, and a login that reports a wrong password to someone typing their own name.

**Your `LOWER()` note was the important one, and it holds.** I tested it rather than reasoning about it:

| `lc_ctype` | `LOWER('MÜLLER')` | folds? |
| --- | --- | --- |
| `en_US.utf8` (what `postgres:15-alpine` defaults to) | `müller` | yes |
| `C` | `mÜller` | **no** |

Every username being ASCII is exactly why this never mattered — `LOWER()` folds ASCII identically under every collation. Widening the class is what introduces the dependency. On a `C`-locale database, `users_username_lower_idx` stops collapsing `Müller` and `müller`, so both accounts can exist: the same impersonation shape I'd just spent the character class avoiding, arriving through a different door. With `eqIgnoreCase`, login also degrades to exact-case for those accounts.

The compose template is fine (`postgres:15-alpine`, `en_US.utf8`), but `DATABASE_URL` can point at an operator's own server, so it isn't hypothetical.

The fix is a collation-independent fold — a stored `username_folded` column maintained by the application, with the unique index and lookups moved onto it. That's a migration, and writing one by hand while this PR is rewriting the migration system would conflict for no reason. So **#8 is a draft until this merges**, at which point it becomes one generated migration on the infrastructure you've built here. The character-class and NFC work in it is unaffected.

So the `TODO.md` entry is right to exist, and it turned out to be the more interesting half of the question rather than a footnote to it.

Nothing outstanding from my side on this PR — the fixes and your follow-up both look right to me.

Azumi0:
@the-luap let me know if there is anything more you would like me to change, otherwise please merge it and I will proceed with stage 2 - SQLite implementation

PR #8, opened 2026-09-03
Allow Latin-script usernames, folded in the application

the-luap:
## Description

The UI ships in German and could not hold `müller`. `usernameSchema`'s character class was `^[a-zA-Z0-9_-]+$` — written as A-Za-z and never revisited.

Raised as an open question while reviewing #7. This PR sat as a draft until #7, #9 and #10 landed, because the blocker below needed the generated-migration system they built.

### The blocker, and why it is the whole design

`TODO.md` recorded the real objection: usernames are compared case-insensitively for both uniqueness and login, and **`LOWER()` folds only ASCII when the database's `lc_ctype` is `C`**. Measured:

| `lc_ctype` | `LOWER('MÜLLER')` |
| --- | --- |
| `en_US.utf8` (what `postgres:15-alpine` defaults to) | `müller` |
| `C` | `mÜller` |

While usernames were ASCII that difference could not matter. Widening the charset while the *database* did the folding would have made `Müller` and `müller` one account on some installs and two on others — the same impersonation shape the character class is chosen to avoid, arriving through a different door. `DATABASE_URL` can point at an operator's own server (`docs/adr/0002`), so this is not hypothetical.

**So the fold moves into the application.** `foldUsername` (NFC, then lowercase) is the single definition, `users.username_folded` stores its result, and every lookup and uniqueness check reads that column. `storage.ts` derives it on create and on rename, so no caller can write a row whose fold disagrees with its username. Uniqueness now means the same thing on every deployment.

### The character class

Latin script plus `0-9`, `_` and `-`. Covers every umlaut, eszett and accent the German and English UIs produce.

**Deliberately not `\p{L}`.** That admits Cyrillic `а` (U+0430), which renders identically to Latin `a` — so `аdmin` and `admin` would be two accounts indistinguishable wherever a username is shown, including the public sharing page, which needs no login to view. Nothing here detects confusables, and folding cannot help: the two differ in every comparison available. Widening to all scripts is a one-token change, but it should be deliberate and paired with confusable detection. Digits stay `0-9` for the same reason — `\p{N}` would admit Arabic-Indic digits.

### Normalisation is part of the rule

`ü` is either U+00FC or `u` + U+0308 depending on the keyboard, the OS, and whether the text was pasted. Left alone they are different strings, so two accounts could display the same name, and someone who registered with one spelling would be told their password was wrong when they typed the other.

It runs *before* the length and character checks, which matters twice: `max(30)` means 30 characters rather than 30 code points, and a decomposed umlaut is not read as a letter followed by a stray combining mark.

Every entry point is covered because #7 routed them all through `storage.getUserByUsername` — registration, `check-username`, login, and both admin endpoints are one chokepoint. The client register form mirrors the rule, normalisation included.

### Migration `0005` refuses rather than corrupting

Two cases it cannot resolve, both checked before anything is altered:

1. **A database whose collation cannot lowercase the non-ASCII usernames it already holds.** Those names could only have been created by an admin, before either `/api/users` endpoint validated anything — which #7's ADR records as a real upgraded-install state.
2. **Two accounts that would fold to one name.** Reachable: the old `lower(username)` index never collided two Unicode spellings, because they are genuinely different strings.

Both name what is wrong and change nothing, so fixing the cause and starting again retries. A refusal leaves `0001..0004` applied and `0005` pending, so a restart retries only `0005`.

### Known limitation

`ß` does not fold to `ss`, so `Straße` and `Strasse` remain two accounts. Recorded as a deliberate limit rather than left to be discovered: full case folding has other one-to-many mappings to decide about, and no database-side rule would agree with it.

### Note on `seed-demo-data.ts`

Users are now written through raw SQL, for the reason the `materials` insert already is: `verify-upgrade.ts` runs the seed against a *pre-migration* database, and drizzle spells every insert with the full column list — it would name `username_folded` before `0005` adds it. Leaving the column out is also what exercises the backfill. `DEFAULT` is used wherever a row does not set a column, so each seeded account keeps the defaults it had as a drizzle insert.

## Type of change

Every username valid before is still valid, and no stored username changes.

## How Has This Been Tested?

- [x] `npx vitest run` — 11 files, **298 tests** passing (287 before, 11 added)
- [x] `npm run check` — clean
- [x] `npx tsx scripts/verify-upgrade.ts` — **all 8 checks pass** with `0005` in the chain
- [x] `0005`'s guards exercised directly against real containers: the duplicate-fold refusal fires on composed-vs-decomposed `müller`, the collation refusal fires on a `--locale=C` database holding `Müller`, both leave `username_folded` unadded, and a healthy database backfills `Alice → alice`, `müller → müller`
- [x] Accepted: `müller` in both encodings, `Straße`, `Jörg-Müller`, `renée`, existing ASCII names
- [x] Refused: Cyrillic `аdmin`, Greek, Han, spaces, `.`, emoji, zero-width joiner, RTL override

The held-name tests from `82200e5` move to a name the rules **still** refuse — `müller` is legal from here on, so a dot stands in for it. The two cases they used to pin become the behaviour change: a name with diacritics can now be recased, and another user still cannot be renamed onto it in any spelling.

## Checklist:

- [x] I have made corresponding changes to the documentation — `TODO.md` entry closed with the reasoning; the rationale lives above `foldUsername` and in `0005`
- [x] I have added tests that prove my fix is effective
- [x] Any dependent changes have been merged and published — #7, #9 and #10


Azumi0:
The argument `UI ships in German` is not entirely true, because UI ships in German **and English**, so the product is aimed at international audience, not just german-based one.  I think it would be better to keep the usernames without diacritics, like e-mail addresses have. 
I would rather suggest implementing another field called "displayName" if you want the user to see their name with proper diacritics like `müller` or `przemysław`.

Azumi0:
Also, because pr #7 reworks migrations and refactors db schema, it would be easier and better to implement such change after the #7  is accepted

the-luap:
## Holding this as a draft — it depends on the database's collation

Azumi0 raised on #7 that usernames are compared with `LOWER()` and that its behaviour on non-ASCII depends on the database's collation. I tested it rather than reasoning about it, and it's a real gap in this PR:

| `lc_ctype` | `LOWER('MÜLLER')` | folds? |
| --- | --- | --- |
| `en_US.utf8` (the compose template's default) | `müller` | yes |
| `C` | `mÜller` | **no** |

Before this PR every username was ASCII, which `LOWER()` folds identically under every collation — so the collation never mattered. Widening the character class is what introduces the dependency.

On a `C`-locale database, with this PR merged:

- **Uniqueness stops holding.** Registration checks `LOWER(username) = LOWER($1)` and `users_username_lower_idx` is `UNIQUE (LOWER(username))`. Neither collides `Müller` with `müller`, so both accounts can exist — two names that differ only in case, side by side wherever usernames are displayed, including the unauthenticated public sharing page.
- **Case-insensitive login degrades.** With #7's `eqIgnoreCase`, someone who registered `Müller` can only log in with that exact capitalisation.

That first one is the same impersonation shape I deliberately kept out by choosing Latin script over `\p{L}`. Letting it back in through the collation would be inconsistent, so this shouldn't merge as it stands.

`docker-compose.template.yml` uses `postgres:15-alpine`, which defaults to `en_US.utf8` and is therefore fine — but Filadex supports pointing `DATABASE_URL` at an operator's own server, and a `C`-locale one is not unusual.

### What it needs

A case-fold that doesn't depend on the server's collation — a stored `username_folded` column maintained by the application (`normalizeUsername(u).toLowerCase()`, which is full Unicode folding in JS), with the unique index and every lookup moved onto it.

That's a schema migration, which is why this is a draft rather than a longer PR: **#7 is actively rewriting the migration system**, and adding a hand-written migration alongside it would conflict for no good reason. Once #7 lands, this becomes one generated migration on the infrastructure it builds.

So the sequencing is: #7 merges → this gets the folded column → out of draft.

The character-class and NFC work in this PR stands on its own and doesn't change; nothing here needs reverting.

### Aside, unchanged either way

`ß` never folds to `ss` under any collation (`LOWER('STRASSE') = 'strasse'`, `LOWER('STRAßE') = 'straße'`), so `Straße` and `Strasse` remain two accounts. Full case-folding would need `ß → ss` handled explicitly, which is a separate decision from this one.

PR #9, opened 2026-09-04
Pay off the TODO.md technical-debt batch

Azumi0:
Works through the Technical Debt section of `TODO.md`, one commit per item, following the plan in `docs/plans/technical-debt-batch.md`. The items are independent; the two commits touching the migration chain come last.

Every commit ends green: `npm test` (248 passing) and `npm run check` both pass, and `npx tsx scripts/verify-upgrade.ts` passes on the final state.

## Commits

1. **Delete the unused `getPublicFilamentsWithUser`** — nothing called it; `GET /api/public/filaments/:userId` does the near-equivalent by composing `getUser` + `getFilaments`, because it must answer 404 where this method threw.

2. **Remove the duplicate `/api/sharing` endpoints** — `POST /api/sharing` and `POST /api/user-sharing` were the same feature with divergent semantics (they diverged once into a bug where sharing could not be switched off). Only `/api/user-sharing` has callers, so `/api/sharing` and its storage method `upsertUserSharing` go. `setUserSharing`'s delete-all-then-insert already collapses the duplicate rows an older release could leave behind, and inherits the comment explaining that hazard. `tests/routes/public.test.ts` drove its setup through `/api/sharing`, so it now uses `/api/user-sharing`; the one test recording the duplicate-row bug is kept and repointed. `docs/API.md` and `docs/TESTING_GUIDE.md` updated to match.

3. **Escape search wildcards instead of stripping them** — `containsIgnoreCase` stripped `%` and `_`, so a term made only of wildcards collapsed to an empty pattern that matched everything. It now escapes `%`, `_` and `\` and adds the `ESCAPE` clause, so those characters match literally.

4. **One password rule everywhere** — `changePasswordSchema.newPassword` required 6 characters while registration and reset require 8, so a user could register with 8 and immediately downgrade. It now uses the shared `passwordSchema`. The rule binds a password being set, not one already stored — short existing passwords still work at login. The two client mirrors are updated to match.

5. **Record the ASCII username rule as deliberate** — documents in `shared/schema.ts` why `usernameSchema` stays ASCII: case-insensitive uniqueness and login agree only while usernames stay ASCII, because `LOWER()` on non-ASCII depends on the deployed database's collation. Replaces the open TODO item with a pointer to that reasoning.

6. **Drop the dead `filaments` timestamp columns** — `filaments.created_at` / `updated_at` were written by `docker-entrypoint.sh` and read by nothing. Generated migration `0002` drops them; `0000` still creates them, so the baseline assertion in `scripts/migrate.ts` is untouched. `scripts/verify-upgrade.ts`'s row-preservation check snapshotted `SELECT *`, which read the dropped columns as mutated data; it now snapshots the columns `shared/schema.ts` defines, still comparing every surviving column value-for-value.

7. **Record the timestamp-column decisions in `TODO.md`** — item 11 (unifying on `timestamptz`) points at `docs/adr/0002-defer-unifying-timestamp-columns.md`; item 10 becomes a done-note explaining the drop above was safe where the general unification is not.

## Verification

- `npm run check` — clean
- `npm test` — 9 files, 248 tests passing
- `npx tsx scripts/verify-upgrade.ts` — all 7 checks pass (Docker)


the-luap:
## Review

I checked each of the seven claims against the code rather than taking them as read, and **all seven hold**. Several were not trivial to verify, so the verdicts are worth stating rather than just listing what's wrong:

- **`0002` really doesn't disturb the baseline guard.** `0000` still declares both columns, so `baselineShape()` still expects them; the legacy chain adds them, `assertMatchesBaseline` passes, the baseline is recorded, *then* `0002` drops. Fresh install is create-then-drop, and a re-run is a no-op because the journal is present. A structural diff of `0001_snapshot.json` against `0002_snapshot.json` shows *only* `public.filaments` losing `created_at`/`updated_at` — no accidental drift. Nothing reads the columns: no triggers, not `FILAMENT_SELECT_COLUMNS`, no client code, and `docker-entrypoint.sh` no longer has a `CREATE TABLE` block to recreate them.
- **The wildcard escaping is correct.** I rendered the emitted SQL through `PgDialect.sqlToQuery` rather than reading it: `"col" ILIKE $1 ESCAPE '\'` with parameter `%50\%\_a\\b%`. The single-pass `/[\\%_]/g` avoids the ordering bug where escaping `%` and `_` first would double-escape the backslashes. It's the only ILIKE construction left in `server/`, and all three call sites go through it.
- **The password rule binds a password being set, not one already stored.** `server/routes/auth.ts:285` validates only the incoming body, and `verifyPassword` never re-checks length. Both client mirrors were updated, and the seeded-5-char-password regression test is the right way to pin it.
- **`/api/sharing` genuinely had no callers left**, and it used `authenticate` rather than `requireApiToken`, so token-based integrations never reached it either.
- **The deleted tests are all mirrored.** I checked all seven removed `public.test.ts` cases against `users.test.ts:689-790`. The only one without a direct replacement — "keeps per-material and global settings as separate rows" — is covered indirectly by the surviving `public.test.ts:140-151`, which shares both and asserts both spools come back. No silent coverage loss.

One finding is worth acting on before merge; the rest are small.

## The row-preservation check no longer covers column drops

`scripts/verify-upgrade.ts:55` — **medium.**

`snapshotData` now selects `Object.values(getTableColumns(table))`, i.e. only the columns `shared/schema.ts` *currently* declares — for the before-snapshot as well as the after. Because dropping a column always comes with deleting it from `shared/schema.ts`, check 1 can no longer observe a column drop **at all**.

Concretely: if `0002` had read

```sql
ALTER TABLE "filaments" DROP COLUMN "purchase_price";
```

with the matching `shared/schema.ts` deletion, `npx tsx scripts/verify-upgrade.ts` would still print

```
PASS  upgrade preserved every row unchanged
```

while every user's purchase price was destroyed. Under the previous `SELECT *` that would have failed loudly, which is exactly why the narrowing was needed here.

So this PR's own drop is covered by no data check: `verifyBackfill` only covers the `filament_types` transformation, and checks 2–4 are schema-shape comparisons that a deliberate drop satisfies by construction. The narrowing is understandable — `SELECT *` genuinely did read the dropped columns as mutated data — but it solved that by making the check blind to the whole category rather than to this instance.

Suggestion: snapshot the columns present in the **pre-migration** database, and require a per-migration allowlist of intentionally-dropped columns. Then a drop has to be *named* to pass, rather than inferred from the schema file — which keeps the check's meaning ("every column that existed still holds its value, unless we said otherwise") instead of narrowing it to "every column we still declare".

## Smaller things

**`scripts/verify-upgrade.ts:74` — the next additive migration turns a diff into a crash.** `dataBefore` is taken against the legacy database before `scripts/migrate.ts` runs, but the column list comes from the post-migration schema. The first future migration that ADDs a column to an existing table makes the before-snapshot issue `SELECT "new_col" FROM "filaments"` against a database that doesn't have it yet — the script throws from inside the `try` and reports nothing about any of the six checks. `SELECT *` produced a diff instead of an exception, so this is a regression in failure quality rather than in coverage.

**`server/db/predicates.ts:35` — `ESCAPE '\'` assumes `standard_conforming_strings = on`.** The generated text is literally `ILIKE $1 ESCAPE '\'`; with that GUC off, the backslash escapes the closing quote and every community-filament search 500s. Genuinely marginal — it has defaulted to `on` since PostgreSQL 9.1 — but `ESCAPE E'\\'` is one backslash regardless of the setting, and `docs/adr/0002` in this same PR argues Filadex supports being pointed at an operator's own database whose settings are unknown. Cheap consistency with the argument the PR itself makes.

**`client/src/i18n/locales/{en,de}.ts:91` — orphaned key stating the superseded rule.** `auth.passwordRequirements` ("Password must be at least 6 characters" / "…mindestens 6 Zeichen…") had exactly two call sites and this PR replaced both with `auth.passwordTooShort`. It's now dead *and* wrong, so whoever reaches for it next ships the old rule.

**`shared/schema.ts:213` — "one password rule everywhere" isn't quite everywhere.** `resetPasswordSchema.newPassword` is still a hand-written `z.string().min(8, ...)`, sixty lines below the `passwordSchema` constant this commit introduces for that purpose. Identical today; the failure mode is the next change to `passwordSchema` silently not applying to password reset.

**`docs/adr/0003-per-user-material-catalog.md` and the two `docs/plans/per-user-material-catalog-*.md` — out of scope.** 456 lines for an unrelated feature in a PR titled "Pay off the TODO.md technical-debt batch", and none of the commit headlines mention them. Docs-only so there's no runtime risk, but it makes the PR harder to review and the ADR numbering harder to attribute later.

**`docs/API.md:1133` — a documented endpoint removed with no deprecation note.** `GET`/`POST /api/sharing` were in the public API reference. There's no CHANGELOG in the repo and the README is untouched, so a third party who scripted the JWT flow against the documented endpoint gets a 404 with nothing pointing them at `/api/user-sharing`. Low because it needed a session/JWT rather than an API token, so the print-server integration surface is unaffected — but the migration note costs a line.

## Checked and clear

- `getPublicFilamentsWithUser` had no references outside docs, and removing it left no orphan imports.
- `setUserSharing`'s delete-then-insert isn't transactional, so a failed insert leaves no sharing row — but it fails toward *private*, and it's pre-existing behaviour this PR only moved a comment onto.
- `ORDER BY 1` in `snapshotData` stays deterministic: every table declares `id` first, so the explicit select list's first column is still the primary key.
- `TODO.md` bookkeeping is uneven — three resolved items got "done/decided" annotations while three others were deleted outright — but that's cosmetic.

Azumi0:
Thanks for the thorough pass. All seven verification points check out against the code here too. Addressed in `fe3f9da..c6082b0`:

**`verify-upgrade.ts` row check blind to column drops (medium)** — `fe3f9da`. The column list now comes from the *pre-migration database*, minus an explicit `INTENTIONALLY_DROPPED` allowlist (`filaments: created_at, updated_at`). A drop has to be named to pass; an unnamed one lands in the snapshot as a `dropped without an INTENTIONALLY_DROPPED entry: …` line and fails check 1 with the column named. Verified by removing the allowlist entry and confirming check 1 goes `FAIL` (cleanly, not a throw). Deriving from the live DB also fixes the smaller finding — the next additive migration no longer makes the before-snapshot `SELECT` a not-yet-existing column and crash the whole run.

**`ESCAPE '\'` and `standard_conforming_strings`** — `4c73cd4`. Now `ESCAPE E'\\'`, one backslash regardless of the GUC. Emitted SQL confirmed via `PgDialect.sqlToQuery`; the wildcard-escaping tests still pass.

**`resetPasswordSchema.newPassword` hand-written `min(8)`** — `4c3cf45`. Routed through `passwordSchema`. Same commit drops the orphaned `auth.passwordRequirements` locale key (dead + stated the old 6-char rule).

**`docs/API.md` `/api/sharing` removal with no migration note** — `c6082b0`. Added a rename note pointing at `/api/user-sharing`, including the request-body change.

`npx tsx scripts/verify-upgrade.ts` and the full `vitest` suite (248 tests) pass.

**On the out-of-scope docs (`0003` + the two `per-user-material-catalog` plans):** these are part of a larger plan drafted in one refinement session, and I'll be implementing them in the next PR — so they shouldn't hurt sitting in the repo now. Leaving them in this branch deliberately.

**`setSharing` delete-then-insert not transactional** — noted; pre-existing, fails toward private, leaving it.

**`ESCAPE E'\\'`'s marginal `standard_conforming_strings` sibling and the whole-table-drop gap in verify-upgrade** — the table-drop case still throws rather than failing cleanly; left a comment saying so, since nothing in play drops a table.

the-luap:
Checked `fe3f9da..c6082b0`. All four land, and I ran the negative test rather than reading the description.

**`fe3f9da`** — emptying `INTENTIONALLY_DROPPED` and re-running gives exactly what you describe:

```
FAIL  upgrade preserved every row unchanged
PASS  upgraded schema matches a fresh install
PASS  upgraded schema matches shared/schema.ts
PASS  fresh schema matches shared/schema.ts
PASS  running the migration again changes no row
PASS  running the migration again changes no schema
PASS  the filament_types backfill preserved every spool's identity
Some checks failed.                                    exit 1
```

with `-- filaments: dropped without an INTENTIONALLY_DROPPED entry: created_at, updated_at` in the after-snapshot. The other six still running is the part that proves it fails cleanly rather than throwing out of the `try`.

Deriving the shape from the live database also fixes the additive-migration crash for free, and it is a bit better than what it replaced: `information_schema` covers tables that exist in the database but are not in `shared/schema.ts`, which the schema-derived list could not see.

**`4c73cd4`, `4c3cf45`, `c6082b0`** all check out — `ESCAPE E'\\'` emits one backslash regardless of the GUC, `resetPasswordSchema` routes through `passwordSchema`, the orphaned key is gone from both locales, and the API reference now points at `/api/user-sharing` with the body change noted.

## One gap left, minor and not merge-blocking

The allowlist is one-directional. `INTENTIONALLY_DROPPED` is read in exactly one place — the `continue` in `discoverShape` — and nothing ever checks that a listed column was actually dropped.

So an *unnamed* drop fails the check, which is the case you set out to close, but a *stale* entry silently excludes a still-existing column from the row comparison, permanently and quietly. Concretely: if `0002` were reverted while its entry stayed, `filaments.created_at` and `updated_at` would keep existing and keep holding data, and no run of `verify-upgrade` would ever compare them again — including a later migration that corrupted them.

Closing it is a few lines after the upgrade: assert each listed column is genuinely absent from the post-migration database, and fail if one is still there. That also turns the allowlist into something self-cleaning — an entry that stops being true gets flagged rather than quietly widening the blind spot.

Not worth holding the PR for, and equally fine as a follow-up or a `TODO.md` line.

## The two you declined

Both fine by me. The out-of-scope docs were a reviewability nit rather than a defect, and if `0003` and the two plans are the next PR's design there is no reason to move them now. `setUserSharing`'s delete-then-insert being non-transactional is pre-existing and fails toward private, which is the right direction to fail in.

Documenting the whole-table-drop case as still throwing rather than failing cleanly is the right call too — a known limitation written down beats one discovered later.

Azumi0:
Closed the one-directional allowlist in `de03e52` rather than deferring it — it was a few lines and the blind spot is the kind that gets found late.

`staleDropEntries` reads `INTENTIONALLY_DROPPED` against the upgraded database and fails on anything still present, so an entry narrows check 1 only for as long as it is true. Ran the negative case you'd want:

```
PASS  upgrade preserved every row unchanged
FAIL  every column the upgrade was allowed to drop is gone
  still present, so the INTENTIONALLY_DROPPED entry excludes live data: filaments.name
PASS  upgraded schema matches a fresh install
PASS  upgraded schema matches shared/schema.ts
PASS  fresh schema matches shared/schema.ts
PASS  running the migration again changes no row
PASS  running the migration again changes no schema
PASS  the filament_types backfill preserved every spool's identity
Some checks failed.
```

with `"name"` temporarily added to the `filaments` entry — the other seven still running, same as your `fe3f9da` test. Unmodified, all eight pass.

the-luap (approved):
Approving. All seven original claims verified against the code, and every finding from the review is addressed:

- `fe3f9da` — the row-preservation check can see a column drop again. Verified by emptying `INTENTIONALLY_DROPPED` and confirming check 1 fails cleanly with the columns named.
- `de03e52` — the allowlist now proves itself. Verified by adding `filaments.name` and confirming check 5 fails while the other seven still run.
- `4c73cd4`, `4c3cf45`, `c6082b0` — escape spelling, password rule, and the API rename note all correct.

Unmodified, all eight checks pass locally and in CI (test, upgrade, build all green on de03e52).

Issue #5, opened 2026-07-07
Docker image starts with unmigrated DB: login 500 due to missing users.role


the-luap:
## Summary

After updating the Docker image from the previous `latest` digest to the current `latest`, existing deployments can no longer log in if the database is still on the pre-RBAC schema. Login returns HTTP 500 because the app queries `users.role`, but the column does not exist. The container's migration runner also fails before applying the required schema changes.

## Image / version

- Image: `theluap/filadex:latest`
- OCI source: `https://github.com/the-luap/filadex`
- Image revision label: `a0c84650edc1c3e0398a8d424575e8e8602a2a16`
- Repo digest: `theluap/filadex@sha256:ef6734e431e50aa4adae11757e1a07551118729024b16f10b184c9f801f3e69b`
- Image ID observed: `sha256:7afb0a10017ac6dfd1ca628e3a95c2d873bca5f76fdf7287f02d6d39c71468a8`
- Previous working digest: `sha256:11b7d694d15fbb121f1e04b01ea9e265cd8257ba5d39c099728e6c2740d793e2`

## Symptoms

After the update:

- `POST /api/auth/login` returned HTTP 500.
- `GET /api/auth/me` returned 401 as expected when unauthenticated.
- Static frontend still returned HTTP 200, so the service looked superficially healthy.

Relevant log excerpts:

```text
Migration failed, but continuing...
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@shared/schema' imported from /app/server/db.ts

[ERROR] Error initializing admin user: error: column "role" does not exist
code: '42703'

[ERROR] Login error: error: column "role" does not exist
code: '42703'
POST /api/auth/login 500 :: {"message":"Server error"}
```

## Expected behavior

An existing deployment should either:

1. Run all required migrations successfully before serving the app, or
2. fail startup loudly if migrations cannot be applied, instead of continuing into a partially migrated state where login fails.

## Actual behavior

The container logs `Migration failed, but continuing...`, starts Express, then login fails because the DB schema is missing columns introduced by the new image, beginning with `users.role`.

## Manual workaround applied

I backed up the DB and manually applied the missing additive migrations directly in PostgreSQL:

- added `users.role`, email verification/reset columns, notification preference columns
- created `email_settings`, `catalog_requests`, `filament_types`, `custom_field_definitions`, `filament_usage_log`, `community_filament_cache`, `api_tokens`
- backfilled `filament_types` from existing `filaments`
[...]

Issue #1, opened 2025-12-13
[FEATURE]


Ferdinand99:
## Problem Statement
I am using Filadex outside of the Eurozone (specifically in Norway, using NOK). currently, the application seems to default to the Euro (€) symbol for all cost-related fields. It is slightly frustrating when tracking filament costs, as the visual representation does not match my actual spending currency. It makes the inventory tracking feel less accurate to my region.

## Proposed Solution
Ideally, this could be implemented in one of two ways:

UI Setting: A field in the "Settings" menu where the user can select their currency from a list (USD, NOK, GBP, etc.) or manually type a custom symbol (e.g., "kr", "$").

Environment Variable: Support for a Docker environment variable (e.g., APP_CURRENCY=NOK) that overrides the default setting on container startup.

## Alternatives Considered
I simply input my local currency values and ignore the € symbol. The math works correctly, but it is visually incorrect. I also checked if the application automatically detected my browser's locale/language settings to adjust the currency, but it appears to be hardcoded to Euro.

## User Value
This would make Filadex much more appealing to the global 3D printing community, rather than just users in Europe. It improves the user experience (UX) significantly by allowing the dashboard and cost calculations to reflect the user's actual real-world context.

## Additional Context
I am running the latest Docker image on Unraid. I really appreciate the modern UI compared to other solutions like Spoolman, and adding localization/currency support would be a great polish to an already great interface.

the-luap:
Thanks for your Input. Maybe I find some time on christmas vacation to implementieren it. I also wanted to redesign the App and make it more Production ready, but lets see.

Ferdinand99:
No problem. Love the product so far, so looking forward to see what you are going to do with it in the future 😊

the-luap:
I have implemented now the function to switch the units and added also the NOK and so on.

PR #10, opened 2026-09-04
Per-user material catalog

Azumi0:
## Summary

Gives every user their own material catalog, layered on top of the shared global one, instead of forcing all users to share a single global material list.

- `materials` gains a nullable `user_id` (owner), with per-catalog uniqueness: unique within the Global Catalog, and unique within each Personal Catalog, case-insensitively. Two *partial* indexes rather than one on `(user_id, lower(name))`, because SQL NULLs never conflict with each other and a single index would let the Global Catalog hold duplicates.
- Filament resolution now goes through one shared path: a declared material (typed on a filament) resolves against the user's personal catalog first, falls back to global, and auto-registers itself in the user's personal catalog if it resolves to nothing — a blank declared material registers nothing. Names are matched case-insensitively and with surrounding whitespace trimmed.
- The CRUD settings factory gained user-scoping as a declared capability, applied to materials; the drying-reminder hygroscopy check and catalog-request approval were updated to match (approval still creates globally; drying reminders are scoped per user).
- Catalog reads now always name whose catalog an entry belongs to; personal entries sort after global ones in the UI, and the settings UI surfaces materials needing attention (unattended) with an explanation on the row, which the owner can dismiss.
- The material migration script refuses to run rather than drop data it can't reconcile, naming every conflicting name.
- Docs (`docs/API.md`, `docs/adr/0003`), i18n strings (`de`/`en`), demo seed data, and tests (`filaments`, `settings`, `notification-checks`, `catalog-requests`) updated to match.
- Dropped `server/types.d.ts`, a stale duplicate of `server/types.ts`'s `Express.Request` augmentation left over from an earlier refactor.

### A name can exist in both catalogs at once

An earlier version of this description claimed personal entries "can't shadow-collide with the global catalog by name". That was wrong, and worth correcting explicitly: the two partial indexes are disjoint, so `(NULL, 'PLA-CF')` and `(alice, 'PLA-CF')` coexist — and they will, whenever an admin approves a Catalog Request for a name the requester had already auto-registered.

That is allowed on purpose. Blocking the approval would let one user's typo veto a shared addition, and collapsing the rows would discard whatever they had filled in. The cost is that *every* read has to break the tie the same way, which is now the case and is written up in `docs/adr/0003`.

## Review findings addressed

Everything from the review at `4ee360f`, one commit per finding:

| # | Finding | Commit |
| --- | --- | --- |
| 1 | A declared material was never trimmed, so `" PETG"` missed the curated global row and auto-registered a duplicate | `dd3b5dc` |
| 2 | `resolveMaterial` was personal-wins while `getHygroscopicMaterialNames` was a flat name-set union, so the two disagreed on a collision | `7fb8b8d` |
| 3 | The migration guard named only one duplicate group per run | `6f83d23` |
| 4 | A case-variant create was a `500`, not a `409` | `85078c7` |
| 5 | `ownsOrIsAdmin` read the `is_admin` mirror rather than `role` | `c47a7e1` |
| 6 | An unparseable density cleared the stored one on the success path | `f33daca` |
| 7 | `PATCH /api/materials/:id/order` was an unscoped write by id | `5b2a071` |
| 8 | `needsAttention` could not be dismissed | `3d57f16` |

Two notes on how #2 and #4 were resolved, since both had more than one defensible answer:

- **#2** is fixed by making the hygroscopy lookup honour the same Personal-before-Global precedence, rather than by blocking the collision at approval time. Blocking would let one user's auto-registered typo veto a Global Catalog addition for everyone, collapsing would discard filled-in data, and neither helps a collision that already exists.
- **#4** adds an opt-in `duplicateOf` capability to the CRUD factory and applies it to `materials` only. The other four entities have case-sensitive indexes, so applying it there would newly *reject* creates that succeed today — a product decision rather than a bug fix. Split out as #11.

Beyond the review: `createMaterial` also normalises the name, or an admin typing `" PETG"` into the Add Material form would create a Global Catalog row that nothing resolves to.

## Testing

- `npm test` — **287 passing**, 11 files (275 before this branch's review fixes, 12 added for the findings above)
- `npm run db:verify-upgrade` — all checks pass with `0004` in the chain, including drizzle-kit's own `pushSchema` diff against `shared/schema.ts`
- `npx tsc --noEmit` — clean
- `npm run build` — clean
- The `0003` guard executed against a real Postgres holding `PLA/pla`, `PETG/petg` and `ABS/abs`: refuses with `(abs, petg, pla)`. Before `6f83d23` the same table produced `(abs)`.

**Not covered by automated tests:** the two client-side fixes (#6, #8). This repo has no client test harness — all 11 test files are server-side — so those are verified by `tsc` and the production build only. Filed as #12.


the-luap:
## Review (at `4ee360f`)

I ran the two things the Testing section omits, since this PR adds a migration: **`npm test` — 275 passing across 11 files**, and **`npm run db:verify-upgrade` — all checks pass**, including drizzle-kit's own `pushSchema` diff against `shared/schema.ts`. `npx tsc --noEmit` clean. So the work verifies better than the description claims; worth listing those two in the PR body, because a reviewer who takes the section at face value would assume neither had been run.

**One thing I expected to find and did not, which is worth stating.** I came in expecting the classic Postgres trap: `UNIQUE (user_id, name)` does not prevent duplicate global rows, because global rows have `user_id IS NULL` and NULLs are distinct in a unique index. You avoided it with two *partial* indexes, and I verified that by execution rather than by reading — applied `0003` to a real Postgres 16 and probed it: a second global row with the same name is rejected by `materials_global_name_lower_idx`, a second personal row for the same user by `materials_user_name_lower_idx`, and two different users may hold the same personal name. The `coalesce(user_id, 0)` comment is accurate about why it's there.

Two findings I'd want addressed before merge. Both are silent data-quality failures rather than crashes, and both were found by running the code, not reading it.

## 1. A declared material is never trimmed

`server/storage.ts:203` (`ensureDeclaredMaterialResolves`) and `:913` (`resolveMaterial`) — **high.**

`eqIgnoreCase` lowercases but does not trim. So `" PETG"` does not resolve to the global `PETG`, and auto-registration then creates a personal row named literally `" PETG"`.

Proven against the PR's own harness: with global `PETG` (density 1.27, hygroscopic) seeded, `createFilament({ material: " PETG" })` left `[" PETG"]` in Alice's Personal Catalog. And `"MoonPLA"`, `" MoonPLA "`, `"MoonPLA "` produced **three separate rows**. The partial unique index cannot catch this — `lower(name)` genuinely differs between them.

What the user sees: the spool silently loses the curated density and the hygroscopic flag, so drying reminders never fire for it (`isOneOfMaterials` compares untrimmed too), and the settings list fills with rows that look identical. A trailing space from a CSV column or a paste is enough.

This is the failure class the ADR exists to eliminate, and nothing currently tests it. Trimming once at the seam — `declared.trim()` for both the resolve and the insert — covers every entry point at once.

## 2. Personal entries *can* shadow-collide with the global catalog, and the two lookups then disagree

`migrations/pg/0003_materials_gains_an_owner.sql:28-29` with `server/storage.ts:513` — **high.**

The description says personal entries "can't … shadow-collide with the global catalog by name". They can: the two partial indexes are disjoint, so `(NULL,'PLA-CF')` and `(1,'PLA-CF')` coexist happily — verified by direct INSERT against the migrated table.

That alone would be defensible. The problem is that the two read paths resolve the collision differently:

- `resolveMaterial` is **personal-wins** ("Personal before Global").
- `getHygroscopicMaterialNames` is a flat **name-set union**, so the global row's `isHygroscopic: true` wins.

Concrete sequence, proven end to end: Alice declares `PLA-CF`, which auto-registers a blank personal row. An admin later approves her catalog request, and approval `createMaterial`s a **global** `PLA-CF` with density 1.24 and hygroscopic true. Now `GET /api/materials` returns both, Alice's spool resolves to her blank personal row, and the drying check reads hygroscopic from the global one. **Alice unchecking "hygroscopic" on her own row has no effect** — she keeps getting reminders and there is nothing in the UI explaining why.

Either make `getHygroscopicMaterialNames` honour the same personal-wins precedence, or collapse/block the collision at approval time. The per-user drying scoping is otherwise correct and tested — this is the one edge the test doesn't reach.

## Medium

**`migrations/pg/0003:19` — the guard names only one duplicate group per run.** `string_agg(DISTINCT lower(name), ', ') … GROUP BY lower(name) HAVING count(*) > 1` aggregates *within* each group, so it's always a single name, and PL/pgSQL `SELECT INTO` without `STRICT` keeps only the first row. Run against a table holding `PLA/pla`, `PETG/petg`, `ABS/abs`, the message was `(abs)` — one name. The comment above it promises "naming the offending rows". An operator with three collisions gets three failed container starts, each naming one. Aggregating over the grouped set fixes it:

```sql
SELECT string_agg(n, ', ') INTO conflict
FROM (SELECT lower(name) AS n FROM materials GROUP BY lower(name) HAVING count(*) > 1) d;
```

Refusing rather than dropping is the right instinct, and it does refuse — this is only about the message.

**`server/utils/settings-crud.ts:153` — a case-variant create is a 500, not a 409.** `POST /api/materials` calls `entityStorage.create` straight after `insertSchema.parse`, with no duplicate check. Before this PR `UNIQUE(name)` was case-sensitive so `"petg"` alongside `"PETG"` succeeded; now the partial `lower(name)` index rejects it and the 23505 falls into the generic catch as **`500 {"message":"Failed to create material"}`** (proven). Every install seeded by `docker-entrypoint.sh` ships `PLA/PETG/ABS/TPU`, so an admin retyping one in the Add Material form gets a 500. The CSV path already does the `equalsIgnoreCase` check correctly — the single-create path just needs the same.

**`server/utils/settings-crud.ts:74-76` — authz reads the mirror, not the source of truth.** `ownsOrIsAdmin` gates on `req.user?.isAdmin`, while every other gate in the codebase is `requireRole("admin")` reading `req.user.role` — and `shared/schema.ts:11-13` documents `role` as the source of truth with `is_admin` as a backward-compatibility mirror. This is now the only authorization decision reading the mirror. I could not find a path that desynchronises them today, so this is prevention rather than a live bug: if one ever appears, `is_admin = true, role = 'user'` grants a non-admin delete and edit rights over Global Catalog rows. `req.user?.role === "admin"` costs nothing.

## Low

- **`client/src/components/settings/settings-materials.tsx:41-49`** — browsers report `""` for a `type="number"` input they can't parse (append `e` to `1.24`). `onBlur` sees `"" !== "1.24"` and PUTs `density: null`, clearing a stored density on the *success* path — no toast, no undo. Guard the blur on the input being in a valid numeric state, not just on the text having changed.
- **`server/utils/settings-crud.ts:259`** — `PATCH /api/materials/:id/order` is the one materials route that never resolves the id through `getAll(req.userId)`. An admin can rewrite `sort_order` on another user's Personal Catalog row, which isn't in their list and which they can't see. Little impact today since globals sort first regardless, but it's an unscoped write on a now-per-user table.
- **Auto-registration is unbounded, by design.** Every distinct declared string permanently creates a personal row — manual create, edit, CSV and Spoolman import alike — and deletion is blocked while a spool still declares it. The ADR accepts this and "needs attention" mitigates it; flagging only because finding #1 multiplies the row count for free. Related: `needsAttention` can't be dismissed, so a personal material that genuinely has no published density and genuinely isn't hygroscopic stays flagged forever — "never touched" is indistinguishable from "deliberately false".

## Checked and correct

- `ON CONFLICT DO NOTHING` without a conflict target **does** arbitrate partial unique indexes — verified against the migrated table for both indexes, so the auto-registration race is genuinely handled.
- `ALTER TABLE materials DROP CONSTRAINT "materials_name_key"` is safe on both paths: `0000` names it explicitly, and the legacy fixture's `name TEXT NOT NULL UNIQUE` gets that exact auto-generated name from Postgres. No `IF EXISTS` needed.
- The migration is additive and doesn't disturb the baseline — `assertMatchesBaseline` parses `0000` only, the journal chain is consistent, and no `INTENTIONALLY_DROPPED` entry is needed since `0003` drops a *constraint*, not a column, so check 5 is unaffected.
- **Blank declared material really registers nothing** — `""` and `"   "` both no-op, and case-only variants correctly resolve to the existing row without creating one.
- **Tenancy is correct on every path I probed**: user A gets 404 on B's personal row for DELETE and PUT, 403 on a global row, and A's filaments cannot resolve against B's catalog. `tests/routes/settings.test.ts` pins all of that plus the four non-scoped entities being unaffected.
- **`server/types.d.ts` deletion loses nothing** — `server/types.ts` carries the same `Express.Request` augmentation and tightens `userId?: number` to `userId: number`, which is what `entityStorage.getAll(req.userId)` relies on.
- `docs/API.md` matches the implemented behaviour, and all seven new i18n keys exist in both `en` and `de`.

Coverage is genuinely good — 130 + 293 new test lines, all passing. The gaps are exactly the three things above: whitespace normalisation, the shadow-collision and its hygroscopy precedence inversion, and the case-variant 500.

Azumi0:
All eight addressed, one commit each, pushed as `dd3b5dc..3d57f16`. I reproduced every finding with a failing test before fixing it — six of the eight failed exactly as described, which is how I confirmed the reproductions rather than taking them on trust.

| # | Commit | Note |
| --- | --- | --- |
| 1 whitespace | `dd3b5dc` | trimmed at the seam, plus one more site — below |
| 2 shadow-collision | `7fb8b8d` | precedence, not blocking — below |
| 3 guard message | `6f83d23` | `(abs, petg, pla)` where it was `(abs)` |
| 4 case-variant 500 | `85078c7` | materials only — below |
| 5 `is_admin` mirror | `c47a7e1` | `req.user?.role === "admin"` |
| 6 density blur | `f33daca` | guarded on `validity.badInput` |
| 7 unscoped order write | `5b2a071` | resolved through `getAll(req.userId)` |
| 8 undismissable flag | `3d57f16` | new `attention_dismissed` column, migration `0004` |

**On #2 — I took the first option and deliberately not the second.** Blocking the collision at approval time would let one user's auto-registered typo veto a Global Catalog addition for everyone, collapsing the rows would discard whatever they had filled in, and neither does anything for a collision that already exists. So `getHygroscopicMaterialNames` now applies the same Personal-before-Global precedence `resolveMaterial` does, and the ADR records that a name living in both catalogs is intended, with the consequence that every read has to break the tie identically. Two tests: the personal row's `false` overriding a global `true`, and the converse still reminding.

**On #4 — scoped to `materials` only.** `manufacturers`, `storage_locations` and `diameters` have case-sensitive `UNIQUE` indexes and `colors` has none, so a case-insensitive check there would newly *reject* `Prusa` alongside `prusa`, which succeeds today. That is a product decision about what those lists mean rather than a bug fix, so it is #11 with the options laid out. The mechanism is an opt-in `duplicateOf` capability on the CRUD factory, so an entity joins with one line. The client now distinguishes the 409 too — a generic "could not be added" toast left the admin no better off than the 500 did.

**One thing your list did not cover, which the same defect implies.** `createMaterial` stored the name untrimmed, so an admin typing `" PETG"` into the Add Material form would create a Global Catalog row that nothing resolves to — `resolveMaterial` compares against the trimmed declared material. Normalising in `createMaterial` covers the Add form, CSV import and Catalog Request approval in one place, which is also why the `duplicateOf` predicate trims: the rule a create has to satisfy is the index *plus* that normalisation.

Your point about the Testing section was fair and the description has been rewritten. It also claimed personal entries "can't shadow-collide with the global catalog by name", which is simply false; that is corrected in its own section rather than quietly dropped.

Verification, all on the final tree:

- `npm test` — **287 passing**, 11 files (275 + 12 new)
- `npm run db:verify-upgrade` — passes with `0004` in the chain
- `npx tsc --noEmit`, `npm run build` — clean
- The `0003` guard executed against a real Postgres holding `PLA/pla`, `PETG/petg`, `ABS/abs`

**One gap, stated rather than buried:** #6 and #8 are client-side and this repo has no client test harness, so they are covered by `tsc` and the build alone. #6 is the one that bothers me — `validity.badInput` is browser behaviour jsdom does not implement for number inputs, so a component test would have reported a pass on the broken code. Filed as #12, arguing for Playwright against the testcontainer the server suite already starts.

the-luap (approved):
Approving. All eight findings verified fixed against the code at `3d57f16`, not just the summary table:

- **#1** `catalogName()` is applied at all five seams — resolve, auto-register insert, both sides of `isOneOfMaterials`, and `createMaterial`. Not folding case there is right, since the database already does via `LOWER()`.
- **#2** the personal-wins tie-break in `getHygroscopicMaterialNames` is correct regardless of row order — the partial indexes guarantee at most one personal and one global row per key. Taking precedence over blocking was the right call, for the reasons you gave.
- **#3** aggregates over the grouped set. **#5** reads `role`. **#7** resolves through `getAll(req.userId)`, consistent with DELETE/PUT.
- **#4** the opt-in `duplicateOf` capability, and deferring the other entities is correct — a caseless check there would newly reject creates that succeed today, which is a product decision rather than a fix.
- **#8** `0004` is purely additive, the journal is sequential, and it needs no `INTENTIONALLY_DROPPED` entry.

Verified locally: `npm test` 287 passing / 11 files, `tsc --noEmit` clean. CI green, and I confirmed the `upgrade` job genuinely ran all 8 checks with `0004` in the chain rather than only reporting success.

The ninth defect you found yourself — `createMaterial` storing an untrimmed name, so an admin's " PETG" would create a global row `resolveMaterial` could never match — was a path my review missed. Good catch.

Noting your own flag on #6: `validity.badInput` is browser behaviour jsdom does not implement, so no component test would have caught the broken version. #12 is the right place for that.

PR #14, opened 2026-09-05
Support SQLite alongside Postgres (stage 2 of 2)

Azumi0:
## Description

Finally, after 3 pull requests, I'm submitting the one that introduces a feature that I wanted from the start :D

This is **stage 2 of two**: adding SQLite as an alternative database engine alongside PostgreSQL. Filadex previously required a PostgreSQL server, which has been the single largest obstacle to single-container homelab deployments tracking one person's filament. With this change, setting `DATABASE_URL=file:/data/filadex.db` yields a complete, zero-external-dependency deployment.

**PostgreSQL remains the supported default.** SQLite is a documented option for single-user/homelab installs. There is no migration path between engines in either direction. However, both engines run the entire test suite in CI — the tier distinction is a deployment/support promise, not a compromise on testing or correctness.

The complete architecture, design trade-offs, and verified evidence are documented in `docs/adr/0004-sqlite-alongside-postgres.md`.

---

### Key Architectural Decisions

1. **Build-time dialect selection & dual server bundles**
   - The database dialect is resolved at compile time via TypeScript path aliases (`@shared/columns` resolves to `columns.pg.ts` or `columns.sqlite.ts`).
   - A single Docker image carries both `dist/index.pg.js` and `dist/index.sqlite.js` produced by `esbuild`.
   - `docker-entrypoint.sh` selects the appropriate bundle and migration runner based on the URL scheme in `DATABASE_URL` (`file:` / `sqlite:` vs PostgreSQL).
   - This keeps Drizzle's compile-time type inference intact across the entire application without introducing an artificial runtime abstraction layer over Drizzle.

2. **Client is built once, with schema invariance proven by tests**
   - The frontend imports `insertFilamentSchema` as a runtime value (`filament-modal.tsx`).
   - Rather than building two client bundles, the client is built once. A new test (`tests/schema-invariance.test.ts`) verifies that all 13 `createInsertSchema` definitions are functionally and structurally identical across both PostgreSQL and SQLite dialects.

3. **Accurate SQLite column semantics & corrections from earlier assessments**
   - **`numeric` maps to `text()` on SQLite:** SQLite's `NUMERIC` affinity converts strings on write and returns JavaScript numbers on read (e.g., `"0.10"` returns as `0.1`), silently breaking string precision for money and weights. Mapping `numeric` to `text()` ensures byte-identical round-trips. All comparisons/aggregations in the application already happen in JavaScript via `Number(...)`.
   - **`t.pk` uses `AUTOINCREMENT`:** Prevents SQLite rowid reuse of deleted spools, ensuring deleted IDs are never reassigned.
   - **Timestamps as epoch milliseconds:** `integer({ mode: 'timestamp_ms' })` avoids sub-second truncation flakes.
   - **Case-insensitive matching:** Dialect-specific predicates in `predicates.pg.ts` (`ILIKE`) and `predicates.sqlite.ts` (`LIKE ... ESCAPE '\'`). SQLite's ASCII-only case folding is documented and accepted.

4. **Workaround for upstream drizzle-kit index parsing bug**
   - Drizzle-kit splits index key expressions on commas without respecting parentheses, so `coalesce("user_id", 0)` is torn apart into invalid SQL.
   - We introduced `nullableIndexKey(column)`: PostgreSQL retains `coalesce(column, 0)` (required for drizzle-kit introspection and `verify-upgrade`), while SQLite uses the bare column within the partial index `WHERE user_id IS NOT NULL`. All catalog uniqueness and shadowing rules from ADR-0003 hold identically on both engines.

5. **Two migration runners & CI migration drift check**
   - `migrations/sqlite/0000_light_catseye.sql` provides a clean baseline schema for SQLite.
   - `scripts/migrate.pg.ts` preserves the legacy migration chain, advisory locking, and baseline assertions for Postgres.
   - `scripts/migrate.sqlite.ts` runs SQLite migrations via `@libsql/client`.
   - `scripts/check-migration-drift.ts` runs in CI (`npm run db:check-drift`) to guarantee that checked-in migrations fully describe `shared/schema.ts` on both engines.

6. **SQLite connection pragmas**
   - Connections via `@libsql/client` initialize with `journal_mode = WAL`, `busy_timeout = 5000`, and `foreign_keys = ON` (critical for `ON DELETE CASCADE`).

7. **Lean Docker image and startup path**
   - Postgres readiness wait, DB creation, and permission probing moved from shell scripts into `scripts/migrate.pg.ts`.
   - Removed `postgresql-client`, `netcat-openbsd`, `tsx`, `typescript`, and `tsconfig.json` from the production image.
   - `docker-entrypoint.sh` is now ~15 lines.

8. **Unified database seeder**
   - Consolidated three drifted seeder scripts into `scripts/seed.ts` with `--starter` (initial catalog items when `INIT_SAMPLE_DATA=true`) and `--demo` (for dev and CI). Removed redundant `init-data.ts` and `seed-demo-data.ts`.

9. **First-class SQLite database backups admin panel & scheduler**
   - SQLite deployments lack `pg_dump`, so Filadex provides a built-in admin backup UI and scheduler.
   - Backups are generated using SQLite's `VACUUM INTO` via `@libsql/client` for consistent live snapshots.
   - Features: on-demand snapshot creation, direct streaming download, backup retention pruning ("keep last N"), and scheduled backups (daily/weekly).
   - Security: all backup routes are admin-only, paths are strictly contained to `/data/backups/`, and backup endpoints reject requests if running on Postgres.

10. **Documentation and Compose templates**
    - Added `docker-compose.sqlite.template.yml` for single-container deployment.
    - Updated `README.md`, `.env.example`, and `docs/DEVELOPMENT.md` with SQLite configuration and workflow guidance.

## Type of change

- [x] Documentation update
- [x] Code refactoring
- [ ] Performance improvement

## How Has This Been Tested?

- [x] **PostgreSQL test suite:** `npm test` — 15 test files, 307 passed, 7 skipped (using testcontainers `postgres:15-alpine`)
- [x] **SQLite test suite:** `TEST_DIALECT=sqlite npm test` — 15 test files, 313 passed, 1 skipped
- [x] **Type checks:** `npm run check` (Postgres) and `npm run check:sqlite` (SQLite) pass cleanly with 0 errors
- [x] **Migration drift check:** `npm run db:check-drift` passes for both dialects
- [x] **Postgres upgrade verification:** `npm run db:verify-upgrade` passes all checks (legacy catch-up, schema equivalence, byte-for-byte data preservation, idempotent re-run)
- [x] **Full build:** `npm run build` cleanly builds Vite client assets and dual esbuild bundles for server, migrator, and seeder
- [x] **CI matrix:** `.github/workflows/test.yml` runs type check and test suite for both `postgres` and `sqlite` matrix legs, as well as migration drift and upgrade verification

## Checklist:


the-luap:
## Review (re-verified at `7267064`)

The review ran against an earlier head and you pushed four commits while it was going, so every finding below was re-checked against `7267064` before posting. None of them were fixed by `88d5973`, `67d0cf4`, `b880eff` or the `main` merge — the two commits named "Remediate … review findings" closed different things.

The architecture is sound and unusually well evidenced. `docs/adr/0004` is the best entry point and it holds up. Most of what follows is not about the dialect abstraction, which is good — it is about what happens at the edges when the URL and the bundle disagree.

Two findings were **reproduced in built images**, and they share one root cause and one cheap fix.

## 1. An unrecognised `DATABASE_URL` scheme silently becomes a Postgres database named after your file path

`docker-entrypoint.sh:5` — **high.**

```sh
case "${DATABASE_URL}" in
  file:*|sqlite:*)  ... sqlite bundle ;;
  *)                ... postgres bundle ;;   # everything else
esac
```

With `DATABASE_URL=sqlite3:/data/filadex.db` and the `PG*` variables the compose template already ships:

- `pg-connection-string` parses it to `{ host: "", database: "data/filadex.db" }`
- the empty host falls back to `PGHOST` / `PGUSER` / `PGPASSWORD`
- `migrate.pg.ts:261` treats `3D000` (database does not exist) as success
- `:275-307` runs `CREATE DATABASE "data/filadex.db"`
- migrations apply, the app boots 200, the default admin is created

**Not one warning.** `FILE:`, `SQLITE:`, `libsql://`, a leading space, and a bare `/data/filadex.db` all take this arm.

The person this hurts is the one this PR is for: someone moving an existing install to SQLite, who typos the scheme or copies a `libsql://` example. They get a fully working, completely empty Filadex, and their real data is simply not there. Nothing in the logs says which database was used.

It is also a regression in blast radius rather than a pre-existing wart: on `main` the shell created `$PGDATABASE`, so a bogus URL still landed on the *correct* database.

```sh
case "${DATABASE_URL}" in
  postgres:*|postgresql:*) ... ;;
  file:*|sqlite:*)         ... ;;
  *) echo "Unrecognised DATABASE_URL scheme: ${DATABASE_URL}" >&2; exit 1 ;;
esac
```

## 2. A relative `file:` path writes the database outside the volume

`server/db.sqlite.ts:26` — **high.**

`.env.example:15` and `README.md:71` both teach `DATABASE_URL=file:./dev.db`. In a container that resolves against `process.cwd()` = `/app`.

Reproduced with `-v fdxdata:/data`: `/data` is empty, `/app/data/filadex.db{,-wal,-shm}` exist, the app is healthy, the admin is created, and no log line names the location. Every byte lives in the container's writable layer and is destroyed by the next `docker compose down && up`, or any image update.

Still current at `7267064` — the block creates the parent directory but never checks `path.isAbsolute` and never logs where it landed.

Both #1 and #2 come down to the same gap: nothing validates that the URL and the running bundle agree, and nothing says out loud which database was opened. Logging the resolved absolute path at startup, and refusing a non-absolute `file:` path when `NODE_ENV=production`, closes #2 and makes #1 obvious the moment it happens.

## 3. `nanoid` is imported at runtime but is not a declared dependency

`package.json:35` — **high, and a one-line fix.**

`server/vite.ts:5` and `server/routes/backups.ts:5` both `import { nanoid } from "nanoid"` at module top level. It is `external` in all six bundles and resolves only through `tailwindcss-animate → tailwindcss → postcss → nanoid@3.3.8`.

It survives `npm ci --omit=dev` *today*. But `main`'s Dockerfile ran `npm install --save-dev tsx typescript` after the prune, which reinstated that tree; this PR removes that step. A transitive-dependency bump, or dropping `tailwindcss-animate`, now produces a GH#5-shaped `ERR_MODULE_NOT_FOUND` at container start. Worth declaring it explicitly.

## 4. The schema-invariance test cannot see the divergence it exists to catch — and one exists today

`tests/schema-invariance.test.ts:42` — **high.**

`describeZod` ends with `return { type: typeName }` for anything it does not special-case, so `ZodString` and `ZodNumber` never have `_def.checks` read. It also drops `ZodDefault._def.defaultValue()` and does not traverse `ZodPipeline`, so `usernameSchema` serialises as `{"type":"ZodPipeline"}` and `passwordSchema` as `{"type":"ZodString"}`.

drizzle-zod already gives a PG `integer` `.min(-2147483648).max(2147483647)` and a SQLite `integer` `±(2^53-1)`. So **every integer column already differs across dialects and line 79's `toEqual` passes.**

`docs/adr/0004:72-81` cites this test as what makes the single client bundle safe, and `vite.config.ts:24` hard-codes the client to `columns.pg.ts`. The next width difference — a `varchar(n)` against a `text` — ships a client that rejects input its own server accepts, with the test still green. Minor: `:74` asserts `> 15` keys when the real count is 41.

## Medium

**`server/backup-scheduler.ts:75` — scheduled backups can still overlap.** `b880eff` closed the due-since gap, and `isBackupDue:26` now refuses when `lastBackupAt` is under an hour old, but there is no in-flight guard: `setInterval` fires regardless, and `lastBackupAt` is only written after `createBackupFile()` resolves. On a daily schedule the stored `lastBackupAt` is ~24h old for the entire duration of a run, so a `VACUUM INTO` that outlives the interval lets the next tick start a second one on the same database, plus a racing `pruneBackups`. A module-level `let running = false` is enough.

**`scripts/verify-upgrade.ts:439` — `scripts/seed.ts --demo` has no automated coverage at all.** verify-upgrade no longer invokes the seeder; it uses an inlined raw-SQL duplicate at `:272-410`. CI runs `npm test`, `db:check-drift` and `db:verify-upgrade`, and none of them run `npm run db:seed` — where `main`'s upgrade job ran `seed-demo-data.ts` on every build. Concretely, `scripts/seed.ts:115` destructures `[admin, alice, bob, unverified]` from a multi-row `.returning()`, which relies on an ordering SQL does not guarantee and nothing now checks. The two fixtures encode the same data twice and can drift apart silently.

**`shared/schema.ts:534` — `updateBackupSettingsSchema` has no bounds.** It is `createInsertSchema` minus `id`/`updatedAt`. The UI clamps retention to ≥1 (`settings-backups.tsx:265`); the API does not. `retentionCount: 0` makes the next `POST /api/admin/backups` write a file and then `pruneBackups(dir, 0)` → `slice(0)` deletes **every** backup including the one just created, while returning 201 with a filename that no longer exists. A negative value makes `slice(-n)` delete the n *oldest* per run. `schedule`, `time` and `dayOfWeek` are equally unvalidated.

**`server/storage.ts:171` — the one SQL-side operator on a `numeric`→`text` column.** `eq(filamentTypes.diameter, diameter)` compares `numeric` on Postgres (`'1.750' = '1.75'` is true) and TEXT on SQLite (false). A CSV import declaring `1.750` reuses the existing filament type on Postgres and creates a duplicate on SQLite. Same root cause at `shared/schema.ts:386`: `diameters.value` is `numeric UNIQUE` on PG, which rejects `1.750` alongside `1.75`, and `text UNIQUE` on SQLite, which accepts it — and `routes/settings.ts:142-166` has no JS-side guard for diameters the way it does elsewhere. ADR-0004:106's "Nothing in the application uses it" wants narrowing to "nothing except this equality".

**`server/storage.ts:990` — non-ASCII catalog duplication on SQLite.** SQLite's `LOWER()` folds ASCII only, so `resolveMaterial` does not match `Äbs` against `äbs`, and `onConflictDoNothing` (`:219`) does not suppress it either, because `materials_user_name_lower_idx` also sees them as distinct. A second identical-looking catalog row appears and the spool inherits neither the density nor the hygroscopic flag. Meanwhile the manual add-material form uses JS `equalsIgnoreCase` (`routes/settings.ts:85`), which *does* fold `Ä`→`ä` and answers 409 — so the two paths disagree on the same question. ADR-0004:149-171 frames this as a search nuisance; for `materials` it is a data-quality one, and it is the failure class ADR-0003 was written to eliminate.

## Lower

- **`.github/workflows/test.yml` — nothing ever applies `migrations/sqlite/0000_light_catseye.sql`.** `tests/helpers/db.ts:83` builds the schema from `shared/schema.ts` through drizzle-kit's API, and `db:check-drift` only compares snapshot JSON. So the one artifact every real SQLite deployment executes is never executed in CI. I applied it by hand and it is fine today — 16 tables, 11 indexes — but a future edit that breaks it ships green and fails on first `docker run`. Postgres has `db:verify-upgrade` for exactly this; SQLite has no equivalent.
- **`server/routes/backups.ts:203` — the snapshot download leaks a descriptor and can take the process down.** `stream.pipe(res)` with no `error` listener and no `stream.destroy()` in `cleanup`. A cancelled download unlinks the temp file but leaves the `ReadStream` open until GC, and an `error` on a listener-less stream is an unhandled `'error'` event, which terminates Node. `stream.pipeline` handles both.
- **`server/routes/backups.ts:31` — prune has no filename-prefix guard, and `BACKUP_DIR` is undocumented.** `listBackupFiles` enumerates *any* `.db`/`.sqlite` file and `pruneBackups:56` unlinks whatever falls outside the newest N. `BACKUP_DIR` (`:14`) is honoured in production but appears in no `.env.example`, README or compose template — while `README.md:119` says backups live "at `/data`". An operator who sets `BACKUP_DIR=/data` has the live `filadex.db` counted as a backup: it permanently occupies a retention slot, and once its mtime is no longer among the newest N it is deleted.
- **`scripts/seed.ts:34` — starter sample filaments dropped and `db:init` lost its env gate.** The four sample spools `init-data.ts` created under `INIT_SAMPLE_DATA=true` were not carried over; `seedStarter` seeds catalog rows only. Docker behaviour is unchanged, so this only affects `npm run db:init` — which now has no `INIT_SAMPLE_DATA` check (`:226-232`) and writes rows where it previously did nothing. Also gone: the `Dark Gray #545454`, `Red #C12E1F` and `Blue #0A2989` colours and the `Keller` location.
- **`docker-entrypoint.sh:29` — extra argv is silently dropped.** The managed branch matches on `$1`/`$2` only and then `exec node "${APP}"` without `"$@"`, so `node dist/index.js --inspect` loses `--inspect` without a word.

## Checked and correct

These took the most work to establish, so they are worth stating rather than leaving implied:

- **No GH#5 regression.** Image built and run on both dialects. `tsx`, `typescript`, `tsconfig.json` and `psql` are all absent from the production image and nothing shells out to them; all six bundles and both `migrations/` directories are present; esbuild inlines the path aliases before `--packages=external` applies.
- **`migrate.pg.ts` lost nothing the shell did.** Readiness wait verified live — Postgres stopped, migrator started, Postgres restarted at 5s, recovered on attempt 5, exit 0. Database creation, permission probing and seed idempotency all preserved, and the advisory lock is a net improvement over the shell version.
- **The SQLite pragma concern is unfounded.** `@libsql/client@0.18.0` on a `file:` URL is a single native handle with no pool, so a per-connection pragma cannot be missed; `foreign_keys=1` verified across 20 parallel executes, inside transactions and inside batches. Cascade verified against the real migrated schema: deleting a user zeroes `filaments`, `filament_types` and `filament_usage_log`. That is load-bearing, since `storage.ts:520` `deleteUser` does no manual child cleanup.
- **`VACUUM INTO` is the right call.** It captures WAL-resident commits — verified that a backup contains a row a naive single-file copy misses.
- **Backup path containment is sound.** `basename` equality, explicit `..` / `/` / `\` rejection, an extension allowlist, and a `resolve` + `sep` prefix check. Express decodes `..%2f` before the check, so the encoded bypass fails. All five routes carry `authenticate, isAdmin, requireSqliteDialect`, and `88d5973` removed the DELETE route entirely.
- **`nullableIndexKey` really is equivalent.** Verified against sqlite3 with the checked-in DDL: global duplicates rejected, same-user duplicates rejected, the same name in two users' catalogs allowed, and the global/personal shadow ADR-0003 deliberately permits still permitted. `ON CONFLICT DO NOTHING` does fire against a partial unique index.
- **Case-insensitive matching matches across dialects.** `eqIgnoreCase` is `LOWER() = LOWER()` on both, with no bare `=`; the email-case-mismatch scenario does not reproduce. `%`, `_` and `\` are escaped single-pass in the correct order on both sides — searching `%` does not widen, and a trailing `\` does not error. Usernames are entirely JS-side (`foldUsername` + `username_folded` on a plain-column index), with no DB-side `LOWER()` anywhere, which is what makes them dialect-proof.
- **The skips are benign.** 307+7 = 313+1 = 314 either way: the 7 Postgres skips are 6 SQLite backup-route tests plus 1 scheduler end-to-end, and the 1 SQLite skip is the Postgres-refusal test. All genuinely dialect-specific.
- **`verify-upgrade` is intact.** All 8 checks byte-identical, `INTENTIONALLY_DROPPED` and `staleDropEntries` untouched, and `:280-288` still omits `username_folded`, so migration `0005`'s backfill is still exercised.
- **The `numeric`→`text` argument holds.** No `ORDER BY`, `sum()`, `avg()`, `GROUP BY` or `<`/`>` over any numeric column exists anywhere in `server/`, so the lexicographic-sort and wrong-aggregate failure modes genuinely do not exist here. The equality in the medium section is the one gap.

---

Nothing here argues against the approach. #1, #2 and #3 are what I would want fixed before this ships, and between them they are perhaps twenty lines.

Azumi0:
All thirteen findings are fixed, pushed as `f961fbe`. Everything below was verified after the change, not asserted.

| Check | Result |
| --- | --- |
| `TEST_DIALECT=sqlite npm test` | 337 passed, 1 skipped (was 324/1) |
| `npm test` (Postgres) | 329 passed, 9 skipped — 338 both ways |
| `npm run check` / `check:sqlite` | clean |
| `npm run db:check-drift` | both dialects match |
| `npm run db:verify-upgrade` | all 8 checks pass |
| `npm run build` | all six bundles + client |

## The two you reproduced in images, reproduced again in a built image

`docker-entrypoint.sh` now names the two schemes it accepts and refuses everything else. Every URL from the finding, against a fresh build:

```
'sqlite3:/data/filadex.db'   exit=1  Unrecognised DATABASE_URL scheme
'libsql://example'           exit=1  Unrecognised DATABASE_URL scheme
'FILE:/data/x.db'            exit=1  Unrecognised DATABASE_URL scheme
' file:/data/x.db'           exit=1  Unrecognised DATABASE_URL scheme
''                           exit=1  DATABASE_URL is not set
```

`-v fdxdata:/data -e DATABASE_URL=file:./dev.db` now dies at migration time instead of booting empty:

> `DATABASE_URL must name an absolute path in production. "file:./dev.db" resolves to "/app/dev.db", which is inside the container rather than a mounted volume, so the database would be lost on the next restart. Use file:/data/filadex.db.`

With the correct URL, `/data` holds `filadex.db`, `-shm` and `-wal`, `/app` holds no database, and the resolved absolute path is logged by the migrator, the seeder and the app — three lines, deliberately, because that is what an operator checks to confirm the volume took.

`nanoid` is declared (`^3.3.18`). In the production image after `npm ci --omit=dev` it resolves, sits at the top level of `node_modules`, and `tsx`, `typescript` and `psql` are all still absent — no GH#5 regression.

## The invariance test

You were right that it was blind, and about the scale: I diffed the two legs with a deepened `describeZod` before changing anything — 276 differing lines, every one the integer bound, and nothing else in the schema differs at all.

It now compares checks, `ZodDefault` values, both halves of a `ZodPipeline`, enum members and literals. The width difference is the one real divergence, so it is held out of the equality and asserted on the property that matters: Postgres, whose schemas the client bundles, must be the **narrower** side. A stricter client rejects input the server would take, which is visible; the reverse ships input the server rejects. Only the two known widths are exempt, so a column growing to `bigint` fails and has to be decided deliberately.

To confirm it is not merely green for a new reason, I temporarily made `t.text` a `varchar(64)` on Postgres — exactly your `varchar(n)` against `text` — and it fails. Key floor is 41, not 15.

## Everything else

Each of these has a test that fails without the fix; I checked that rather than assuming it.

- **Scheduler** — module-level in-flight guard. Without it the new test hangs and times out.
- **Backup settings** — enum, `HH:MM` regex, `dayOfWeek` 1–7, `retentionCount >= 1`. `pruneBackups` also floors at 1, because a row written by an earlier build can still hold 0. Verified live in the image: `PUT {"retentionCount":0}` → `400 Number must be greater than or equal to 1`.
- **`storage.ts:171`** — new `eqNumeric` predicate, casting on the SQLite side so both engines behave the way Postgres always did. The diameter catalog gained the matching JS guard, so `1.750` against `1.75` is a 409 on both instead of a 500 on one and an accepted duplicate on the other. ADR-0004:106 narrowed to name this equality.
- **Non-ASCII `LOWER()`** — folded in JS via `foldMaterialName`, `foldUsername`'s rule applied to a trimmed name. The `resolveMaterial` test fails on SQLite with the old SQL version.
- **Prune** — restricted to the `filadex-backup-` prefix. A test drops a foreign `.db` in `BACKUP_DIR` and asserts it is neither listed nor deleted. `BACKUP_DIR` is now in `.env.example`, the compose template and the README, which no longer says backups live at `/data`.
- **Snapshot download** — `stream.pipeline`, with a client abort separated from a server fault so it does not log an error for a cancelled download.
- **CI** — `seed-sqlite` and `seed-postgres` jobs, each fixture against its own database (they collide if chained: `--demo` refuses a populated `users` table and its catalog inserts are plain inserts). `migrate.sqlite.ts` applies `0000_light_catseye.sql`, so that DDL is now executed in CI. The demo fixture no longer destructures a multi-row `.returning()` positionally. `db:migrate:sqlite`, `db:init:sqlite` and `db:seed:sqlite` were added — `@db` resolves to Postgres without the SQLite tsconfig, so a local SQLite dev could not run the seeders at all.
- **Entrypoint argv** — `/proc/1/cmdline` in a running container reads `node dist/index.sqlite.js --unknown-flag-xyz`.

I could not reach a published container port on this machine (`-p` yields `{"5432/tcp":[]}`), so the Postgres seed job was verified inside the container's own network namespace against a real `postgres:16-alpine`: `waitForDatabase` reports "target database does not exist yet", `migrate.pg.ts` creates `filadex_demo`, and `--demo` seeds all fifteen tables.

## Three places I did not follow the review literally

- **Non-ASCII materials — fixed in the application, not with a folded column.** The `username_folded` treatment means a schema change and a backfill on both engines, which is a large thing to add to this branch. Folding in JS closes the failure you described and the disagreement with the 409 path. The residual, stated in the ADR rather than hidden: the partial unique indexes still fold ASCII only, so a concurrent double-declare of `Äbs`/`äbs` on SQLite could still slip past `onConflictDoNothing`. If you want the index to enforce it too, that is a separate change.
- **`db:init` was not re-gated on `INIT_SAMPLE_DATA`.** On `main` that flag only ever gated the sample data, and the README documents `db:init` as a setup step that fills the dropdowns — gating it would make the documented setup silently do nothing. The four sample spools are restored on the flag instead, called from `main()` so `seedStarter`'s early return does not skip them.
- **`Keller` was not restored.** The three Bambu Lab colours are back. Replacing `Keller` with "Dry box A"/"Shelf" alongside translating the rest of the list into English reads as a deliberate rewrite, so undoing it did not seem mine to make.

## One thing found while checking the Synology guide

Not from this review, but on the branch: `docs/deployment-synology.md` pointed Moonraker at `/api/v1/spool`. That is not a route — it falls through to the static handler and returns the web interface's HTML with a 200, so the failure is a parse error with no hint. The endpoints are under `/api/spoolman-compat/v1/`. Fixed in `7dd398a`, along with the expected startup log and the backup filename pattern.

the-luap:
## Second round — the remediation commits only (`93df32f`)

All thirteen findings from the first round are fixed; I checked each against the code rather than the summary table, and independently reproduced the invariance test failing when `t.text` becomes `varchar(64)` on Postgres. The "Postgres must be the narrower side" framing is better than what I suggested — a stricter client rejects input the server would take, which is visible, where the reverse is not.

This round covers only what nobody had reviewed: the three commits pushed in response (28 files, +1,268/−142). The rebase itself is clean — no conflict markers, and #8's `username_folded` / migration `0005` / `foldUsername` are intact and not duplicated. Both legs pass on the head: `npm test` 329/9 skipped, `TEST_DIALECT=sqlite npm test` 337/1 skipped.

Six findings. The first two are the reason this round was worth running.

## 1. `docker-compose.synology.yml:60` — a working, publicly-known JWT signing key ships to users

**High. Confirmed.**

```yaml
- JWT_SECRET=REPLACE-WITH-YOUR-LONG-RANDOM-SECRET-HEX-32
```

That is a literal value, not a placeholder anything rejects. `server/auth.ts:12` is `if (process.env.JWT_SECRET) return process.env.JWT_SECRET;` — no length check, no placeholder check, no warning.

The guide tells the user to paste this file into Container Manager verbatim. The stack comes up healthy, and every install that did so signs session cookies with a string published in this repository. `jwt.sign({ userId: 1 }, "REPLACE-WITH-YOUR-LONG-RANDOM-SECRET-HEX-32")` is a valid admin session on all of them.

The sharp part is that `resolveJwtSecret`'s own comment already says exactly this:

> Falling back to a fixed, publicly-known string would let anyone forge tokens for any unconfigured deployment, so an unset `JWT_SECRET` instead gets a random per-process secret.

So **leaving it unset is strictly safer than this file's default**, and the other two compose files get it right — `docker-compose.template.yml:41` and `docker-compose.sqlite.template.yml:24` both use `${JWT_SECRET:-}`, which lands on the random-per-process path. Only the new Synology file introduces the weak default. Matching the other two is the fix.

Secondary: `docs/deployment-synology.md:170` shows a *different* placeholder (`PASTE-YOUR-OWN-RANDOM-32-CHAR-SECRET-HERE`) inside a block that also says to copy the repo file, so the two disagree about what the user is meant to replace.

## 2. `scripts/seed.ts:93` — sample spools land in a real inventory on the second container start

**Medium-high. Reproduced.**

The docblock at `:88-91` justifies the design with "The Docker entrypoint seeds before the application runs, so there is nobody to own them there and this is a no-op." That holds for start #1 only. `docker-entrypoint.sh:51` runs the seeder on **every** start when `INIT_SAMPLE_DATA=true`, and `docker-compose.synology.yml:70` ships `INIT_SAMPLE_DATA=true` with the comment "Safe to leave 'true' on fresh install".

Sequence: deploy on Synology → the app creates the default admin (`server/auth.ts:155`) → the user logs in, changes the password, adds nothing yet → the NAS reboots → the seeder now finds an `admin` row and an empty `filaments` table, and inserts four fake spools ("PLA Black Bambu Lab", "PETG Transparent", …) into that person's real inventory. The same happens to anyone who empties their inventory and later restarts.

Reproduced directly: a second `db:init:sqlite` against a database that already had an admin printed `Sample spools inserted: 4.` The guard is `filaments` count > 0, which cannot tell "fresh install" from "install the user has not put anything into yet". Gating on the admin's `created_at`, or on a marker row, would separate the two.

## 3. `server/db/predicates.sqlite.ts:48` — `eqNumeric` makes every unparseable diameter equal on SQLite

**Medium. Confirmed — and a regression introduced by the fix for the earlier finding.**

```ts
sql`CAST(${column} AS REAL) = CAST(${value} AS REAL)`
```

`CAST(x AS REAL)` yields `0.0` for anything SQLite cannot parse, and otherwise stops at the first non-numeric character. Verified against libsql:

| comparison | result |
| --- | --- |
| `'abc' = 'xyz'` | **1** |
| `'' = '0'` | **1** |
| `'1.75mm' = '1.75'` | **1** |
| `'1.750' = '1.75'` | 1 *(the intended fix)* |
| `'1.75' = '1.8'` | 0 |

`shared/schema.ts:316` types `diameter` as `z.union([z.string(), z.number()]).nullable().optional()` with no numeric validation, so a CSV import, a Spoolman import or an API-token client can supply `1.75mm` or `""`. `findOrCreateFilamentType` (`server/storage.ts:171`) then matches that spool onto an existing filament type with a *different* diameter, and the spool silently inherits it.

The plain `eq` it replaced compared strings and did not do this — so this is new, and it is the same class of cross-engine divergence `eqNumeric` was added to remove (Postgres errors on the parameter rather than coercing). A `typeof()`/`GLOB` guard in the predicate, or validating diameter as numeric at the schema, closes it. The schema-level fix is probably the better one, since it also stops `1.75mm` reaching the catalog at all.

## 4. `server/routes/settings.ts:185` — the diameters delete guard was left on exact string equality

**Low-medium. Confirmed.**

The same hunk introduces `sameDiameter` and wires it into `duplicateOf` and the CSV `parseLine`, but `isInUse` stays as:

```ts
isInUse: (filament, item) => filament.diameter === String(item.value)
```

With `1.750` in the catalog (imported before this change, or created on SQLite where `text UNIQUE` allowed it alongside `1.75`) and spools carrying the filament type's `1.75`, `"1.75" === "1.750"` is false. `DELETE /api/diameters/:id` therefore sees no usage and removes a diameter that spools are actually using — the one thing that guard exists to prevent. Both sides want `sameDiameter`.

## 5. `tests/schema-invariance.test.ts:27` — the `WIDTH_BOUNDS` comment promises a tripwire the code does not have

**Low. Confirmed.**

The comment says "a column that grows to bigint produces bounds outside this set and fails, which is the point." Checked against drizzle-zod: a Postgres `bigint("b", { mode: "number" })` produces `min: -9007199254740991, max: 9007199254740991` — both values *are* in `WIDTH_BOUNDS`. So a column widened to bigint is silently exempted from the equality comparison and passes the bound assertion.

Nothing dangerous ships in that specific case, because SQLite's integer produces the same pair and Postgres is therefore not the wider side. But the guard advertised in the comment is not there, and the exemption keys on the bare numeric values rather than on the column actually being an integer-width bound. Either correct the comment or key the exemption to the declared type.

## 6. `server/routes/backups.ts:114` — the new prefix restriction covers listing and pruning but not download

**Low. Plausible.**

`listBackupFiles`, and therefore `pruneBackups`, now enumerate only `filadex-backup-*.db|.sqlite`, which correctly stops retention deleting an operator's own files. `resolveBackupPath` did not get the same filter, so `GET /api/admin/backups/:filename` still resolves any `*.db`/`*.sqlite` in `BACKUP_DIR`. With `BACKUP_DIR=/data` — the misconfiguration the new comment at `:14-20` explicitly names — an admin can download `filadex.db`, the live database, although it is not listed. Admin-only, and an admin can already stream a snapshot, so the impact is small; the two halves should just use one predicate.

## Checked and clean

- **`server/sqlite-url.ts`** handles `file:`, `file://`, `sqlite:`, `?mode=ro`, and returns null for `file::memory:` with or without params. No false positive in the production absolute-path refusal: every `DATABASE_URL` in the Synology compose, the Synology guide and `.env.example` passes both the entrypoint `case` and the absolute-path check. The `NODE_ENV=test` bypass only substitutes `file::memory:` for a non-SQLite ambient URL, so it cannot let a bad production value through.
- **`eqNumeric` on Postgres** is genuinely equivalent to the `eq` it replaced, and the new "1.750 reuses 1.75" test passes on the PG leg. All three `equalsIgnoreCase` call sites were migrated.
- **`foldMaterialName`** is applied to *both* sides of every comparison — `resolveMaterial`, the `getHygroscopicMaterialNames` winner key, `isOneOfMaterials`, and both `duplicateOf`/`isInUse` for materials. The JS fold is strictly coarser than SQL `LOWER()` (it adds NFC), so it can only turn a would-be 500 into a 409, never the reverse. Non-defect note: `resolveMaterial` now reads the whole in-scope material set per call, i.e. once per row of a bulk import — small by construction, but no longer index-served.
- **Scheduler guard** — `running` is cleared in a `finally` wrapping both the settings read and the backup, and `createBackupFile` failures are caught inside, so no path leaves it stuck. The new test exercises release-and-rerun.
- **Backup validation** — the `HH:MM` regex requires zero-padding and rejects `25:00`/`2:00`; `dayOfWeek` 1-7 matches both `isBackupDue`'s ISO mapping and the client's `SelectItem` values, so no previously-valid payload starts 400ing. `pruneBackups`'s `Math.max(1, Math.floor(...))` genuinely handles a legacy row holding 0 or a negative, and `NaN` short-circuits harmlessly.
- **`filadex-backup-`** does match what `createBackupFile` produces — both derive from the same constant.
- **The pipeline change** separates client-abort from server-fault on `res.headersSent`, cleans up in `finally`, and leaves no unhandled rejection.
- **`scripts/seed.ts` row identification** is deterministic — a name-keyed `Map` with a throwing lookup, no positional `.returning()`. All four fixture paths run against SQLite: `--starter` idempotent, `--demo` refuses a populated database on the second run.
- **The two new CI jobs assert by failing** — the seeders throw on a missing schema, so a broken `0000_light_catseye.sql` cannot exit 0, and `db:migrate:sqlite` does apply it. `seed-postgres` reaches its service container and `migrate.pg.ts:291` creates `filadex_demo` on `3D000`.
- **Entrypoint passthrough** — `shift 2` + `exec node "$APP" "$@"` is correct for zero args and all three entrypoint spellings.
- **Synology compose, apart from finding 1** — the port is bound to `127.0.0.1`, no database port is published, `./data:/data` is the documented persistent path, and the container runs as root matching the Dockerfile, so there is no permissions trap.

---

Findings 1 and 2 are the ones I would fix before merge: both reach users who follow the new guide exactly, and neither is visible to CI. 3 and 4 are a pair — the diameter value is either validated as numeric or it is not, and right now the predicate, the schema and the delete guard each assume something different.

Azumi0:
## Second round remediation — `4765472`..`40a21ce`

All six findings verified against the code before anything was changed, then fixed in five commits, one per finding. Both legs pass on the head: `npm test` 335/9 skipped, `TEST_DIALECT=sqlite npm test` 343/1 skipped (338 → 344 total; six new tests, each confirmed to fail with its fix reverted).

Two things came out differently from what was suggested. Details below.

### 1. JWT signing key — fixed, but not with `${JWT_SECRET:-}` (`4765472`)

Confirmed, including that `resolveJwtSecret` takes any non-empty string with no length or placeholder check.

The line is now commented out rather than matching the other two compose files. Container Manager takes pasted YAML with no `.env` to interpolate from, so `${JWT_SECRET:-}` would resolve to empty on every Synology install and the guide's "replace the placeholder" instruction would point at nothing. Commented out, an unedited paste lands on the same random-per-process path, and the line the guide tells you to edit is still there to edit.

The doc's second, different placeholder is gone — `docs/deployment-synology.md` now shows the same commented line, and says not to reuse a secret printed in this repository. No placeholder secret string remains anywhere in the repo.

### 2. Sample spools on restart — fixed (`fc9f858`)

Confirmed, and the reproduction matches.

Gated on `admin.lastLogin` rather than `created_at` or a marker row. An empty `filaments` table cannot tell "fresh install" from "install whose owner has not added a spool yet"; the admin having signed in once can, and it needs no new column or time window. Past that point the fixtures never appear again, whatever the flag says and however empty the table gets. The first Docker start is unaffected.

### 3. `eqNumeric` coercion — fixed at the write boundary, not the schema (`72e7af3`)

Reproduced your table exactly against libsql: `'abc' = 'xyz'` → 1, `'' = '0'` → 1, `'1.75mm' = '1.75'` → 1.

You suggested the schema-level fix, and it looked like the better one to me too — but `insertFilamentSchema` is imported only by `client/src/components/filament-modal.tsx`, and that form already sends `diameter` as `z.number()`. `POST /api/filaments` is `const data = req.body` with no parse at all. A refine there would have been inert against exactly the callers you named: CSV import, JSON import, the Spoolman-compatible API, API-token clients.

So `diameterValueSchema` is defined in `shared/schema.ts` (used by the client form, unchanged in spirit) and enforced at the storage chokepoint every write path funnels through — `createFilament` and `updateFilament`. Only caller-supplied values are checked, so a row already holding `1.75mm` stays editable rather than becoming a 400 on an unrelated edit.

`predicates.sqlite.ts` is deliberately unchanged. A GLOB guard there is a second definition of "numeric" to keep in step with the first, and it would not help the rows that motivated it: values already in the database predate the rule either way. Validating at the boundary is the version that has one answer.

### 4. Delete guard — fixed (`72e7af3`)

`isInUse` uses `sameDiameter`. Predicate, JS helper and guard now agree. Covered by a test that deletes a `1.750` catalog entry against a spool carrying `1.75` and expects 400.

### 5. `WIDTH_BOUNDS` comment — comment corrected (`40a21ce`)

Confirmed your check: drizzle-zod gives a Postgres `bigint({ mode: "number" })` ±9007199254740991, both already in the set.

I corrected the comment rather than building the tripwire it promised. Keying the exemption to the declared type would need the describe script to carry Drizzle column metadata it does not currently see, and the property that matters is already asserted separately — Postgres no wider than SQLite. A bigint on both sides passes because neither side is wider; a bigint on Postgres against a narrower SQLite column fails. The comment now describes that instead of a check that isn't there.

### 6. Backup download — fixed (`5a78091`)

`listBackupFiles` and `resolveBackupPath` share one `isBackupFilename`.

One knock-on: the existing download test probed the 404 path with `nonexistent-backup.db`, which the prefix rule now correctly answers 400. Renamed to a prefixed absent name so it still exercises "well-formed name, file missing", and added the case that was missing — a `filadex.db` placed in `BACKUP_DIR` is refused and left on disk.

the-luap (approved):
Approving at `40a21ce`. Two review rounds, nineteen findings, all fixed and all verified against the code rather than the summaries.

**Round two's two headline findings were the reason not to merge on green CI**, and neither was visible to it: a working JWT signing key shipping in the Synology compose, and sample spools landing in a real inventory on the second container start. Both reach a user who follows the new guide exactly.

**Twice you checked my suggestion and correctly declined it.** Both times you were right and I had missed something:

- The JWT fix is commented out rather than `${JWT_SECRET:-}`, because Container Manager pastes YAML with no `.env` to interpolate from — my version would have resolved empty on every Synology install and left the guide pointing at nothing to replace.
- Diameter validation went to the storage chokepoint rather than `insertFilamentSchema`, because that schema is imported only by `filament-modal.tsx` and `POST /api/filaments` never parses it. My suggestion would have been inert against exactly the callers I named: CSV, JSON, Spoolman, API tokens. I verified both.

`WIDTH_BOUNDS` is the other one worth noting: correcting the comment rather than building the tripwire it advertised is the right call when the check would need column metadata the describe script does not carry, and the comment now says what actually holds a widened column honest.

What I verified independently across the two rounds, beyond confirming each fix is present: the invariance test genuinely fails when `t.text` becomes `varchar(64)` on Postgres; `CAST(x AS REAL)` really does make `'abc' = 'xyz'` and `'' = '0'` true on libsql; no placeholder secret string remains anywhere in the repo; `nullableIndexKey` enforces the same rule on both engines including the shadow-collision ADR-0003 permits; and the rebase across #8 left `username_folded`, `foldUsername` and migration `0005` intact.

The architecture was sound from the first read. Almost everything found sat at the edges — what happens when the URL and the bundle disagree, what a fresh container does on its second start, what `CAST` does to a string that is not a number. That is where a dual-dialect port goes wrong, and the evidence you attached to each remediation made checking them fast.

Eight checks green, including both test legs, both new seed jobs, drift, and the Postgres upgrade verification.

PR #24, opened 2026-09-07
Add Polish language support, and make the UI language reach the login screen

Azumi0:
Adds Polish as a third UI language, and fixes the `<html lang>` and pre-login
language-selection gaps that adding it exposed.

## What's here

**Polish locale.** `client/src/i18n/locales/pl.ts` mirrors every key in `en.ts`
(812 keys, placeholders included — `tests/i18n/locale-parity.test.ts` enforces
this against both `de` and `pl`). Server-side: all five email templates, the
`/api/users/language` validation, scheduled notification checks, and the
date-fns picker locale.

**`<html lang>` follows the UI language.** `client/index.html` hard-coded
`lang="de"` and nothing ever updated it, so the document language was always
`de` whatever the user had selected. The server now resolves the language for
the initial HTML (stored preference → `language` cookie → `Accept-Language` →
`en`) and rewrites the tag; the client keeps `document.documentElement.lang`
and the cookie in sync from then on. `docs/adr/0005` records the decision and
the alternatives rejected.

**A language selector before login.** The only selector lived in the header,
which renders behind auth — so a visitor the server's guess got wrong had to
log in through a language they couldn't read. It's now on all five auth
screens. `reset-password` and `verify-email` are included deliberately: they're
reached from an email link, often on a different device from the one holding
the cookie, so they're more exposed to a bad guess than login is.

**Both sides detect the browser language the same way.** `resolveLanguage`
scans the whole `Accept-Language` header for the first supported tag; the
client only read `navigator.language`'s primary subtag and gave up if that one
was unsupported. For a browser set to "French, then German" the server stamped
`de` while the client fell through to `VITE_DEFAULT_LANGUAGE` — the two
disagreed, for anonymous visitors too. The client now iterates
`navigator.languages`, so they agree by construction.

**`DEFAULT_LANGUAGE` removed.** It was set in three compose files and
documented in the README and on Docker Hub, but no server code has ever read
it. A new user's language now comes from the request at registration. ADR 0005
records why it can't simply be wired back up. The build-time
`VITE_DEFAULT_LANGUAGE` is unaffected and is genuinely reachable.

**`shared/languages.ts`.** The supported-language set was spelled out in six
places, one of which carried a "keep in sync with the client" comment. It's now
declared once and imported across client and server.

## Adding a fourth language

`docs/TRANSLATION_GUIDE.md`'s checklist was stale — it pointed at a `resources`
object that doesn't exist and a "settings component" that isn't the selector.
It's been rewritten to walk every real touchpoint.

## Verification

`npx vitest run` — 22 files, 409 tests: 400 passed / 9 skipped on Postgres,
408 passed / 1 skipped under `TEST_DIALECT=sqlite`. The skip counts differ
because the backup cases are dialect-gated; CI runs both. `npm run build &&
npm run test:e2e` — 6 passed. `npx tsc` and `npx eslint . --max-warnings 0`
both clean.

New coverage: server and client language resolution, the production static
catch-all (including that it forwards a rejection to `next()` rather than
hanging the request, its cache headers and 304 revalidation, and the explicit
`/index.html` path), the auth-screen selectors, per-locale notification
dispatch driven off `SUPPORTED_LANGUAGES`, email templates across every
supported language plus escaping of user-supplied values, registration not
inheriting a signed-in account's language, and locale parity.

Two of these are Playwright specs, because the behaviour only exists in a real
browser and `tests/components/*` use `renderToString`, which runs no effects:
`login-language.spec.ts` (a language chosen on `/login` survives a reload and
the login, and reaches the account) and `register-language.spec.ts` (an account
registered while another session is open is stamped from the request, and the
selector on `/register` does not rewrite the signed-in account).

## Notes for review

- The logged-in branch of `resolveLanguage` costs one `getUser` per HTML
  document request. Unauthenticated requests skip it. Called out in the ADR's
  consequences.
- `express.static` is mounted `{ index: false }` so `/` reaches the catch-all
  that stamps the tag. `index.html` is read once at boot, so a rebuild-in-place
  needs a restart — also in the consequences.
- Two dead documentation links (`TRANSLATION_GLOSSARY.md`,
  `TRANSLATION_TECHNICAL.md`) were removed from the README and CONTRIBUTING;
  neither file has ever existed in the repo.

https://claude.ai/code/session_01CqU4ATMkjXwnH95sWeE6pn


the-luap (commented):
Thanks — this is a big, careful PR and the ADR, the shared language set, and the locale-parity test are all the right shape. CI is green and it merges cleanly. I still would not merge it as it stands: four of the findings below are correctness bugs in the feature the PR exists to add. Each was verified against the checked-out branch, not just the diff.

## Must fix before merge

**1. A malformed `language` cookie blanks the whole app** — `client/src/i18n/resolve-language.ts:56`

`getLanguageCookie()` calls `decodeURIComponent` on the raw cookie value with no try/catch, and it runs inside the `useState` initializer of `LanguageProvider` (line 24). There is no error boundary anywhere in `client/src`. A cookie value like `language=%E0` throws synchronously on every route, including `/login`, until the user clears cookies. The client only ever writes plain `en`/`de`/`pl` and the server reads the cookie undecoded, so the decode can simply go.

**2. The new login-screen selector does not survive login** — `client/src/i18n/LanguageProvider.tsx:84`

Picking Polski on `/login` stores it in localStorage and the cookie only, because `setLanguage` skips the POST when there is no `userData`. After login, `/api/auth/me` returns the stored `language` (still `'en'`, the schema default), the `[userData]` effect calls `setLanguageState('en')` *before* consulting localStorage, and the `[language]` effect overwrites the cookie with `en`. The choice this PR just made possible is undone on the next render. Either the effect needs to prefer a fresh local choice over the stored one, or login needs to persist it.

**3. Registration can inherit another account's language** — `server/routes/auth.ts:69`

`POST /api/auth/register` resolves the new account's language via `resolveLanguage(req)`, whose first priority is the DB preference of whatever `token` cookie is present. `/register` is an unguarded route (`App.tsx:86`, no redirect for authenticated users), so user A (language `de`) creating account B stamps B with `de` and sends a German verification mail. Relatedly: `LanguageProvider` still runs the `/api/auth/me` query on `/login` and `/register`, so using the new selector there while a session exists fires `POST /api/users/language` and rewrites the *logged-in* user's preference. Registration should read only the cookie and the header.

**4. The server's `<html lang>` stamp loses to the browser on the client** — `client/src/i18n/resolve-language.ts:33`

`resolveClientLanguage` ranks `documentLang` (priority 5) below `navigator.languages` (3) and `VITE_DEFAULT_LANGUAGE` (4). A user with DB language `pl`, in a fresh profile with `navigator.languages = ['de-DE','de']`: the server stamps `pl`, the client's browser loop returns `de`, and the `[language]` effect rewrites `documentElement.lang` and the cookie to `de`. On `/public/*` the account query is disabled, so it stays German permanently; elsewhere it flashes de → pl. The stamp exists precisely for this case and should rank above the browser list.

## Should fix

- **Every unmatched request now costs a JWT verify plus a full user-row read** — `server/utils/resolve-language.ts:15`. Both SPA catch-alls call `resolveLanguage` for every method and every unmatched path, and there is no `/api` 404 handler in `server/routes/index.ts`, so a mistyped API path, `/favicon.ico`, or a scanner with a valid cookie each trigger `SELECT * FROM users` (password hash included) to pick en/de/pl. The client keeps the `language` cookie in sync on every load and change, so checking the cookie *before* the token, or guarding on `req.method === 'GET' && req.accepts('html')`, gives the same answer with near-zero DB work.
- **The HTML response lost its cache headers** — `server/vite.ts:93`. `main`'s `sendFile` emitted ETag, Last-Modified and answered `If-None-Match` with 304. `res.end(string)` bypasses all of that, and there is no `Vary: Cookie, Accept-Language`. Behind a proxy or CDN that caches `text/html`, the first visitor's `lang="pl"` is served to everyone. `res.set('Vary', 'Cookie, Accept-Language')`, `Cache-Control: no-cache`, and `res.send` (so ETag/304 return) would cover it. Also: `GET /index.html` still bypasses the stamp — `index: false` only affects directory requests.
- **Unannounced behaviour change for self-built deployments** — `client/src/i18n/resolve-language.ts:33`. On `main` only `de` beat `VITE_DEFAULT_LANGUAGE`; now any supported browser language does. A build with `VITE_DEFAULT_LANGUAGE=de` shows English to an `en-US` browser where it used to show German. Narrow reach (the Dockerfile has no `ARG` for it, so Docker builds never had it), but a silent regression for the deployments that do.

## Minor

- `server/utils/email-templates.ts` — each of the five templates grew an `if (pl) … if (de) … return en` chain with an implicit English fallthrough, so a fourth language compiles cleanly and silently mails English. A `Record<Language, …>` per template — the form this PR already uses for `DATE_LOCALES` in `filament-modal.tsx` — makes the compiler catch it. Those templates also interpolate user-controlled `entityLabel`, `reviewNote` and filament names into HTML unescaped; that is pre-existing for en/de, but this adds a third copy rather than fixing it once.
- `LanguageProvider.tsx:89` — the `[userData]` effect re-assembles the same five sources `getInitialClientLanguage()` already builds (with slight drift: `?? null` vs the typeof guard). `setLanguageState(getInitialClientLanguage())` is behaviour-preserving and lets four imports go.
- `server/vite.ts:88` — `fs.readFileSync` at `serveStatic` setup now aborts the whole server, API included, if `dist/public/index.html` is missing; `main` only failed per request.
- `tests/server/serve-static.test.ts` writes and unconditionally deletes a fixture at the real `server/public/index.html` path inside the repo.
- `ResolvedLanguage` and the two `export type { Language }` re-exports have no importers, and `formatSupportedLanguages`'s `languages` parameter plus its 0/1-length branches exist only for a test.

One refuted candidate, for the record: the `setHtmlLang` regex does match the real `client/index.html` today.

Azumi0:
All four must-fix items reproduced and fixed, along with the three should-fix and all five minor items — e296d4d, plus a correction to finding 2's fix in c12ebaf (see the last section; please review c12ebaf rather than e296d4d alone). `npx vitest run` — 22 files, 400 passed, 9 skipped; `npm run test:e2e` — 5 passed; `npx tsc` and `npx eslint . --max-warnings 0` clean; `npx vite build` clean.

## Must fix

**1. Malformed cookie.** `decodeURIComponent` removed from `getLanguageCookie`. The cookie only ever holds a bare language code and the server reads it undecoded, so there was nothing for it to do. Covered by a test that asserts `language=%E0` does not throw.

**2. Login-screen selector.** `LanguageProvider` now holds a choice made with no session and flushes it to `POST /api/users/language` the first time account data appears, taking your "login needs to persist it" branch rather than "prefer the local choice" — preferring `localStorage` unconditionally would have made a language set on device B lose to a stale local value on device A.

**3. Registration inheriting a language.** `resolveAnonymousLanguage(req)` (cookie + `Accept-Language`, no session lookup) is now what `POST /api/auth/register` uses, and also `forgot-password` and `resend-verification` — same class of bug, since those identify the target account by email, not by the browser's session. Pinned by a test that registers with another account's session cookie present and asserts the Polish mail; it fails on the parent commit.

For the client half: `LanguageProvider` now derives its account context from the route, using the `PUBLIC_ROUTES` list that already exists in `client/src/lib/auth.tsx`, so the selector on those five screens cannot write to a signed-in user's preference. One residual: if you pick a language on `/login` while logged out and then land on an authenticated route, that pick is written to whatever account the session resolves to. That is the mechanism fix 2 needs, and I think it is the right behaviour — but it is a deliberate choice, not an oversight.

**4. Stamp vs. browser list.** `documentLang` moved above `browserLanguages`, so the order is `localStorage` → cookie → stamp → browser list → `VITE_DEFAULT_LANGUAGE` → `en`.

## Should fix

**Per-request DB read.** Took the `req.method === 'GET' && req.accepts('html')` guard rather than checking the cookie before the token. Cookie-before-token would let a stale `language` cookie outrank the stored account preference — the one input the server has that the client cannot see, and the input finding 4 just made authoritative on the client. The guard keeps that ordering and drops the lookup for every non-GET and non-HTML request. Two tests assert `storage.getUser` is not called for either.

I did not add an `/api` 404 handler. A GET to a mistyped API path from a browser (`Accept: */*`) still reaches the stamp and still costs the lookup. That is a real gap, but a JSON 404 for unmatched `/api/*` is a behaviour change with a blast radius well outside this PR, so I have raised it separately as #25 — it turns out unmatched `/api` paths answer `200 text/html` for every method today, which is worth fixing on its own terms and not only for the lookup. Happy to take it if you want it.

**Cache headers.** Both serving paths now go through one helper that sets `Vary: Cookie, Accept-Language` and `Cache-Control: no-cache` and sends via `res.send`, so Express computes an ETag and answers `If-None-Match` with a 304 again. `/index.html` is routed to the same handler explicitly, since `index: false` only covers directory requests. Tested, 304 included.

**`VITE_DEFAULT_LANGUAGE`.** This one I could not fix in code, and I want to be explicit about why rather than quietly leave it. Once the stamp outranks the browser list, `VITE_DEFAULT_LANGUAGE` is unreachable wherever the app serves itself: the server always stamps a supported value, and the server's own order already puts `Accept-Language` above anything build-time. There is no ranking for it below the stamp that is also reachable. So the fix is to stop it being *silent*: the README entry now says it ranks lower than it used to, gives your exact `VITE_DEFAULT_LANGUAGE=de` + `en-US` example, and says what to do instead; ADR 0005's consequences say the same. It still applies if the built client is served by something that does not stamp the tag.

## Minor

- Five email templates are `Record<Language, …>` maps — a fourth language is now a `tsc` error, not a silent English mail. `escapeHtml` applied once to `entityLabel`, `reviewNote` and filament names. I escaped `&`, `<` and `>` only: every one of those values lands in text content, never an attribute, and escaping `'` turned `Alice's spool` into `Alice&#39;s spool` in the mail source (it broke an existing notification-checks assertion). The comment on `escapeHtml` says what has to change if a value ever moves into an attribute.
- The `[userData]` effect now calls `getInitialClientLanguage()`; four imports gone, and the `?? null` / `typeof` drift with it.
- `serveStatic` reads `index.html` per request. A missing file now fails that request instead of the whole server, and rebuilding in place no longer needs a restart — that consequence is struck from the ADR.
- `tests/server/serve-static.test.ts` captures any existing `server/public/index.html` and writes it back, and uses `fs.rmSync` for the directory it created.
- `ResolvedLanguage`, both dead `export type { Language }` re-exports, and `formatSupportedLanguages`'s parameter and its 0/1-length branches are gone. (`client/src/i18n/index.ts`'s `Language` re-export stays — `LanguageProvider` imports it.)

## Finding 2, and a correction

I first said finding 2 could not be tested without adding a DOM environment. That was wrong — I checked for jsdom and `@testing-library` and stopped, missing `tests/e2e/`. `tests/e2e/login-language.spec.ts` now covers it (c12ebaf): choose Polish on `/login`, sign in, assert the account ends up on `pl` and that `<html lang>` survives a reload.

Writing it showed **the fix I posted above was wrong**, so the version you review should be c12ebaf, not e296d4d. `AuthProvider` renders a loading placeholder instead of its children while it re-checks the session, and does that on every route change — so the `/login` → `/` navigation unmounts `LanguageProvider` and destroyed the `useRef` I was holding the pending choice in. On screen it looked fixed, because `localStorage` still said `pl`; the account was never told, so the next sign-in elsewhere was English again. The pending choice now lives in module scope, which outlives the remount and still resets on a real page load, by which point the cookie and the stamp carry it. The spec fails against the ref and against the original code.

It signs in as the seeded `alice` (stored `language: "en"`, the schema default the bug turned on) rather than the admin the materials specs use, since the suite shares one database at `workers: 1` and this spec writes account state. Full run: 5 passed.

## Docs

ADR 0005 records the new client order and why the stamp outranks the browser list, the pre-login choice being pushed to the account after login, `resolveAnonymousLanguage` and the HTML-GET guard, the cache headers, and the `VITE_DEFAULT_LANGUAGE` consequence. `docs/TRANSLATION_GUIDE.md` step 6 now points at the five `Record<Language, …>` maps and notes that `tsc` enforces them.

https://claude.ai/code/session_01CqU4ATMkjXwnH95sWeE6pn

the-luap (commented):
Reviewed `e296d4d` against the branch head, not just the diff. All four must-fix items are resolved the way described, the should-fix and minor items too, and the reasoning in each place where you took a different branch than I suggested holds up. `npx tsc`, `npx eslint . --max-warnings 0` and CI are clean on the head. I'm fine with this merging.

What I checked, per item:

- **Cookie.** `getLanguageCookie` returns the raw match; the `%E0` test covers the initializer path. Good.
- **Login-screen choice.** The pending-ref mechanism works because `LanguageProvider` stays mounted across login: `AuthProvider` only swaps in its loading screen before the first `checkAuth`, `login()` is a `setUser`, and `login.tsx` navigates client-side. So the ref survives to the first `account` and the flush runs once. Your reason for pushing the choice to the account rather than preferring `localStorage` — device B's choice must not lose to a stale value on device A — is the right call. The acknowledged residual (a pick on `/login` with a session still in the browser lands on that account on the next authenticated route) is the same mechanism and I agree it is the correct behaviour.
- **Registration.** `resolveAnonymousLanguage` on register, forgot-password and resend-verification, and the route-derived `account` on the client. The new test registers with another account's session plus `language=pl` and asserts the Polish subject; that does fail on the parent. Good catch on forgot-password and resend-verification being the same class.
- **Stamp vs. browser list.** Order is now localStorage → cookie → stamp → browser → `VITE_DEFAULT_LANGUAGE` → en, with the reason in the doc comment.
- **Per-request lookup.** The GET + `accepts('html')` guard, with cookie-before-token rejected for the right reason: a stale cookie must not outrank the stored preference the server alone can see. The `/api` 404 gap being #25 rather than this PR is the right scoping — it is a behaviour change on its own.
- **Cache headers.** `Vary: Cookie, Accept-Language`, `Cache-Control: no-cache`, `res.send` for ETag/304, and `/index.html` routed to the same handler ahead of `express.static`. Per-request `readFile` so a missing shell fails one request, not the server.
- **`VITE_DEFAULT_LANGUAGE`.** Agreed that there is no reachable rank below the stamp. Documenting it as a behaviour change rather than pretending otherwise is the right outcome.
- **Templates.** `Record<Language, …>` per template, and escaping `&`/`<`/`>` only, with the attribute caveat in the comment. Fine for text content.

Two small things, neither blocking:

1. `LanguageProvider.tsx` — the `eslint-disable` on the `[account]` effect is avoidable: in TanStack Query v5 (`^5.60.5` here) `mutate` is referentially stable, so `[account, updateLanguageMutation.mutate]` satisfies `exhaustive-deps` without re-running the effect on every render.
2. The PR description still says "9 skipped"; the head runs 409 tests with 1 skipped (the existing backups case). Worth a one-line edit so the description matches.

For the record: my local full SQLite run had one failure, in `catalog-requests.test.ts`'s `beforeEach` — `verify-email` answering 401 inside `registerAndVerify` — and three reruns of the file passed. That is the register → verify → login helper family that #19 and #21 both noted, not this commit; the route this PR touched has no 401 path at all.

the-luap (commented):
Reviewed `c12ebaf` on the branch head. `npx tsc`, `npx eslint . --max-warnings 0`, `npm run build` and the five e2e specs all pass locally; CI is green. The correction is right, and I owe one of my own: in the previous review I said the ref would survive because `AuthProvider` only shows its placeholder before the first `checkAuth`. That was wrong — its effect runs `checkAuth` on every `location` change and `setLoading(true)` swaps the children out, so `/login` → `/` does unmount `LanguageProvider`. Good catch, and the right instinct to distrust a fix with no test.

What I verified:

- **The spec pins the bug.** With `e296d4d`'s ref version of `LanguageProvider` swapped in and the bundle rebuilt, `login-language.spec.ts` fails with `Expected: "pl", Received: "en"` on the stored preference. With `c12ebaf` it passes. So the module-scope variable is what makes it work, not the surrounding changes.
- **Asserting the account before the DOM** is the correct order, for the reason the comment gives — `<html lang>` reads `pl` mid-transition from `localStorage` either way.
- **`alice` rather than `admin`**, and the database is rebuilt per run by the `webServer` command, so the spec's premise (stored `en`) is re-established every time.
- **CLAUDE.md section** — accurate, including the `renderToString` runs-no-effects point that this whole episode turned on.

**One confirmed gap, non-blocking but worth closing while the spec is fresh.** Module scope survives the remount but not a page load, and the comment says the cookie and the stamp carry the choice "anyway" — they carry the *display*, but the `[account]` effect still prefers `account.language` over both, so the account is never told. I probed it: sign in as `bob` (seeded `de`), pick Polski on `/login`, **reload**, then log in → `/api/auth/me` still says `de` and `<html lang>` reverts to `de`. Reload-before-login is not exotic — a failed first attempt, a password manager, or the `/reset-password` link opening in a fresh tab and the visitor picking a language there. `sessionStorage` instead of a module variable fixes it in the same four lines: it survives reload, is per-tab, and dies with the tab, which is the lifetime the comment is describing. A `page.reload()` between the pick and the login in the spec would pin it.

Two things carried over from the last pass, still non-blocking: the `eslint-disable` on the `[account]` effect is avoidable since `mutate` is stable in TanStack Query v5, and the PR description's "9 skipped" is stale (the head runs 409 with 1 skipped).

Fine to merge as is if you would rather take the reload case as a follow-up.

Azumi0:
All three items from the two reviews are in `bbeaa97`.

**The reload gap is real, and the module variable was the wrong lifetime.** I reproduced it before changing anything: adding `page.reload()` between the pick and the login makes `login-language.spec.ts` fail on `c12ebaf` with the stored preference still `en`. Your reading of why is exactly right — the cookie and the stamp restore the *display*, which is what made it look fixed, while the `[account]` effect goes on preferring `account.language` over both, so the account is never told.

It is `sessionStorage` now, wrapped in `try`/`catch` so a browser that refuses session storage loses only the deferred write; the choice still applies to that tab through `localStorage` and the cookie. The reload stays in the spec, and it is what pins this — it fails against the module variable and passes against `sessionStorage`. The ADR now records all three attempts and why the first two lose the choice, since "a ref is fine here" is exactly the assumption that has now been wrong twice.

**`eslint-disable` gone.** `mutate` is wrapped in `React.useCallback` in the installed v5, so it is stable as you said. One wrinkle: naming `updateLanguageMutation.mutate` in the dependency array does not satisfy the rule — it wants the object the member is read from. Destructuring is what actually does it, so the mutation is now `const { mutate: persistAccountLanguage } = useMutation(...)` and the effect depends on `[account, persistAccountLanguage]`. `npx eslint . --max-warnings 0` is clean with no suppression anywhere in the file.

**PR description updated.** Both dialects, since the skip counts are dialect-gated and CI runs both: 22 files, 409 tests — 400 passed / 9 skipped on Postgres, 408 passed / 1 skipped under `TEST_DIALECT=sqlite`. Your 409-with-1-skipped was the SQLite run; the stale "387 passed, 9 skipped" was mine on Postgres from before the new tests. The description also now lists the two browser specs and says why they are browser specs.

On the `catalog-requests.test.ts` flake: `verify-email` answering 401 inside `registerAndVerify` is in the register → verify → login helper family, and this branch does not touch that route's auth path. I did not chase it here. It reproduced for you once in a full SQLite run and passed on three reruns; both my dialect runs above were clean, so I have nothing to add beyond confirming it is not this change.

Current head is `bbeaa97`. `npx vitest run` and `TEST_DIALECT=sqlite npx vitest run` as above, `npm run test:e2e` 6 passed, `npx tsc`, `npx eslint . --max-warnings 0` and `npm run build` clean.

https://claude.ai/code/session_01CqU4ATMkjXwnH95sWeE6pn

the-luap (commented):
Reviewed `01819ff` and `bbeaa97` on the branch head. `npx tsc`, `npx eslint . --max-warnings 0`, `npm run build` and all six e2e specs pass locally; CI is green on `bbeaa97`. Nothing left open from my side — this is ready to merge.

What I verified, with the fixes reverted one at a time against the rebuilt bundle:

- **`sessionStorage` is what closes the reload gap.** With `c12ebaf`'s module-variable provider swapped back in, `login-language.spec.ts` now fails at the account assertion (`Expected: "pl", Received: "en"`) because of the `page.reload()` step; with the head it passes. The `try`/`catch` around both accessors is right — a browser refusing session storage should only lose the deferred write.
- **The register spec's client half is real.** Making the provider ignore the route (`account = userData`, query always enabled) fails the second assertion with `bob` rewritten to `pl`. So the route-derived account context is what protects the signed-in account, not an accident of ordering.
- **The register spec's server half is now covered twice.** Swapping `register` back to `resolveLanguage(req)` does *not* fail the spec on the head — because `resolveLanguage` only consults the session for `GET` + `Accept: html` since `e296d4d`, and registration is a POST. The commit message's "fails against its own half reverted" holds against the pre-`e296d4d` resolver, not the current one. Not a problem — belt and braces — but worth knowing the spec is pinning the guard as much as the anonymous resolver.
- **`persistAccountLanguage` destructuring** satisfies `exhaustive-deps` without a suppression, and the effect does not re-run per render. Good catch that naming the member on the object does not.
- **PR description** now matches both dialect runs.

One observation, not a change request. The residual you called out earlier is now directly exercisable by the register spec's setup: `bob` picks Polski on `/register` to read the form, and that choice sits in `sessionStorage` as pending. The spec never navigates `bob`'s tab to an authenticated route, but if it did, the pending `pl` would be flushed to `bob`'s account — the choice was made to read a registration form for someone else. That is the deliberate behaviour the ADR records, so I am not asking for a change; it is the one case where "a pre-login pick belongs to whoever logs in next" and "the visitor is not editing the account that still has a session" pull in opposite directions, and the ADR might say which one wins there.

PR #30, opened 2026-09-08
Support PUID and PGID in container entrypoint and Synology compose

Azumi0:
## Problem

Commit 43b63a0 updated the production container to drop root privileges and run as user `node` (`1000:1000`) via `su-exec`.

When deploying on Synology NAS (Container Manager / DSM) using a bind-mounted directory (e.g. `./data:/data`), files and folders created on the host are owned by the DSM user (typically `UID 1026`, `GID 100`) and governed by Synology DSM ACLs. 

Because `docker-entrypoint.sh` hardcoded `node:node` (`1000:1000`) and did not run `chmod -R 775`, the entrypoint dropped privileges to UID 1000. Under Synology DSM's filesystem and ACL rules, user 1000 does not have write access to the host-mounted `/data` directory or existing database files. This caused the container to enter an infinite crash loop on startup:

```text
Applying database migrations...
[INFO] SQLite database: /data/filadex.db
Applying SQLite database migrations...
SQLite database migration failed: LibsqlBatchError: SQLITE_READONLY: SQLITE_READONLY: attempt to write a readonly database
  [cause]: SqliteError: attempt to write a readonly database
  code: 'SQLITE_READONLY',
  rawCode: 8
```

## Solution

1. **Configurable PUID / PGID in `docker-entrypoint.sh`:**
   - Read `PUID=${PUID:-1000}` and `PGID=${PGID:-1000}`.
   - When starting as root (`id -u == 0`), take ownership with `chown -R "$PUID:$PGID"` and enforce `chmod -R 775` on the data and backup paths.
   - Drop root and re-execute via `exec su-exec "$PUID:$PGID" "$0" "$@"`.
   - Log `Using PUID: $PUID, PGID: $PGID` on container startup.
   - Added an automated pre-flight directory writability check before `node "${MIGRATOR}"` runs to provide clear, actionable troubleshooting steps (pointing to DSM File Station permissions / PUID) rather than an unhandled driver crash.

2. **Synology Compose Template (`docker-compose.synology.yml`):**
   - Added `- PUID=1026` and `- PGID=100` to the environment block (matching DSM 7's default user/group UID).

3. **Documentation (`docs/deployment-synology.md`):**
   - Updated the *Permissions Note* in Stage 4 and the Compose template in Stage 5 to explain `PUID` and `PGID`, how to inspect them via `id <username>`, and how to set File Station permissions if ACL issues persist.

## Verification

- `npm run check` -> clean
- `npm run check:sqlite` -> clean
- `npm run lint` -> clean (0 errors, 0 warnings)
- Reproduction harness verified:
  - Non-writable path triggers preflight error with clear Synology guidance.
  - Writable path runs migrations cleanly and boots successfully.

### DIFF
--- a/docker-compose.synology.yml
+++ b/docker-compose.synology.yml
@@ -48,6 +48,11 @@ services:
       - NODE_ENV=production
       - PORT=8080
 
+      # UID/GID of the folder owner on NAS. Check via SSH: id your_username
+      # Typically on Synology DSM the first user is 1026, and "users" group is 100.
+      - PUID=1026
+      - PGID=100
+
       # SQLite database URL.
       # MUST be an absolute path starting with file: (or sqlite:).
       # In production, relative paths (e.g. file:./dev.db) are rejected on startup to
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -4,27 +4,68 @@ set -e
 PUID=${PUID:-1000}
 PGID=${PGID:-1000}
 
+# Parse SQLite file path if DATABASE_URL points to a file.
+# Strips file: or sqlite: prefix and any query parameters.
+# Only absolute paths (starting with /) set DB_DIR, avoiding unintended
+# operations on the application tree for :memory: or relative paths.
+DB_FILE=""
+DB_DIR=""
+case "${DATABASE_URL}" in
+  file:*|sqlite:*)
+    DB_FILE="${DATABASE_URL#file:}"
+    DB_FILE="${DB_FILE#sqlite:}"
+    DB_FILE="${DB_FILE%%\?*}"
+    case "${DB_FILE}" in
+      //*) DB_FILE="${DB_FILE#//}" ;;
+    esac
+    case "${DB_FILE}" in
+      /*) DB_DIR="$(dirname "${DB_FILE}")" ;;
+    esac
+    ;;
+esac
+
 # The image runs the server as an unprivileged user (defaulting to node:node 1000:1000,
 # or PUID:PGID on environments like Synology DSM). Mounted volumes arrive owned by root
 # or the NAS user, so when started as root take ownership of the data paths first,
-# ensure permissions (chmod -R 775), then re-run this script as PUID:PGID. Started as
-# any other user - `docker run --user`, or a compose `user:` line - this block is skipped
+# ensure directory permissions, then re-run this script as PUID:PGID. Started as any
+# other user - `docker run --user`, or a compose `user:` line - this block is skipped
 # and the paths are assumed to be writable already.
 if [ "$(id -u)" = "0" ]; then
-  echo "Using PUID: ${PUID}, PGID: ${PGID}"
-  DATA_PATHS="/data ${BACKUP_DIR:-/data/backups}"
-  case "${DATABASE_URL}" in
-    file:*)
-      DB_FILE="${DATABASE_URL#file:}"
-      DATA_PATHS="${DATA_PATHS} $(dirname "${DB_FILE%%\?*}")"
+  case "${PUID}" in
+    0)
+      echo "ERROR: Running as root (PUID=0) is not supported. Filadex must run as an unprivileged user." >&2
+      exit 1
+      ;;
+    ''|*[!0-9]*)
+      echo "ERROR: PUID must be a valid numeric user ID (got '${PUID}')." >&2
+      exit 1
       ;;
   esac
+
+  case "${PGID}" in
+    ''|*[!0-9]*)
+      echo "ERROR: PGID must be a valid numeric group ID (got '${PGID}')." >&2
+      exit 1
+      ;;
+  esac
+
+  echo "Using PUID: ${PUID}, PGID: ${PGID}"
+  DATA_PATHS="/data ${BACKUP_DIR:-/data/backups}"
+  if [ -n "${DB_DIR}" ]; then
+    case " ${DATA_PATHS} " in
+      *" ${DB_DIR} "*) ;;
+      *) DATA_PATHS="${DATA_PATHS} ${DB_DIR}" ;;
+    esac
+  fi
+
   for p in ${DATA_PATHS}; do
     mkdir -p "${p}"
     chown -R "$PUID:$PGID" "${p}"
-    chmod -R 775 "${p}"
+    chmod u+rwX,g+rwX "${p}"
   done
-  exec su-exec "$PUID:$PGID" "$0" "$@"
+  if [ "$PUID" != "0" ]; then
+    exec su-exec "$PUID:$PGID" "$0" "$@"
+  fi
 fi
 
 # Decide dialect from DATABASE_URL scheme.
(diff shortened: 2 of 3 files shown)


the-luap (commented):
Thanks — the Synology problem is real, and configurable `PUID`/`PGID` is the right shape for it. CI is green and the diff is small, but the entrypoint change has enough defects that I would not merge it as it stands. Each item below was checked against the diff and, where it is shell behaviour, with `sh`.

## Blocking

**1. `chmod -R 775` on every start undoes the backup protection from #29** — `docker-entrypoint.sh:25`

`server/routes/backups.ts` writes every snapshot `0600` and the backup dir `0700` on purpose: each snapshot holds every password hash and pending token. The next container restart now runs `chmod -R 775` over `/data` and `/data/backups`, so all snapshots, `filadex.db` and its `-wal` become `0775` (exec bit included), and nothing ever re-applies `0600` to existing files. With the template's `PGID=100` — the `users` group, i.e. every DSM account — any non-admin user browsing the share can copy the hash dump. The PR also deletes the doc sentence that stated the `0600` guarantee and replaces it with advice to grant **Everyone** Read & Write recursively, so no document now says who can read the database.

The `chown` alone gives the owner what it needs. If a mode fix is wanted, make it `chmod u+rwX,g+rwX` **non-recursively on the directories**, never on the files. (Also: with the default config `DATA_PATHS` expands to `/data /data/backups /data`, so the backups subtree is walked several times per start.)

**2. Every `sqlite:` URL now fails startup with a misleading banner** — `docker-entrypoint.sh:76`

The preflight matches `file:*|sqlite:*` but strips only the `file:` prefix: `DB_FILE` stays `sqlite:/data/filadex.db`, `dirname` yields `sqlite:/data`, `touch sqlite:/data/.filadex_write_test` resolves under `/app` and fails, and the container exits blaming NAS permissions on a perfectly writable volume. `sqlite:` is accepted by the dialect switch at line 45, by `server/sqlite-url.ts`, and is documented as valid in `docker-compose.synology.yml`; it booted on `main`. Strip both prefixes once (`DB_FILE=${DATABASE_URL#file:}; DB_FILE=${DB_FILE#sqlite:}`) and reuse it in both blocks — the root block at line 17 also still adds the DB directory to the chown list only for `file:*`, so a `sqlite:` URL outside `/data` gets told to set PUID/PGID, which that block will never honour.

**3. `PUID=0` loops forever** — `docker-entrypoint.sh:27`

`su-exec node:node` guaranteed the re-exec dropped root. `su-exec "$PUID:$PGID"` with `PUID=0` (the linuxserver.io convention for "run as root", or `id root` pasted in) re-enters the root block, reruns `chown -R` and `chmod -R`, prints `Using PUID: 0` again, and never reaches `node`. The healthcheck marks the container unhealthy with no explanatory error. Refuse `0`, or guard the re-exec with `[ "$PUID" != 0 ]`.

## Should fix

- **Non-root path breaks on a fresh volume** — line 78. The header comment still says `--user` / compose `user:` is supported. In that path nothing runs `mkdir -p`, so `DATABASE_URL=file:/data/db/filadex.db` on a fresh writable `/data` fails the `touch` with ENOENT and the Synology banner. On `main` the migrator created the directory (`server/db.sqlite.ts` `mkdirSync(..., { recursive: true })`).
- **Relative and in-memory URLs chmod the app tree** — line 17. `dirname` of `:memory:` or `./dev.db` is `.` (= `/app`), so the root block runs `chmod -R 775` over `dist/`, `migrations/` and `node_modules/`, and the non-root preflight fails on `/app` with the NAS banner where `main` gave `db.sqlite.ts`'s clear "must be absolute in production" message.
- **`PUID`/`PGID` are interpolated unvalidated** — line 24. `PUID=paul` (the `id your_dsm_username` tip invites exactly this), or `PUID='1026 '` pasted from Container Manager's env editor, makes `chown` fail with a raw busybox error under `set -e` — before any of the new guidance. `case "$PUID" in ''|*[!0-9]*) ... exit 1;; esac` covers it.
- **The preflight proves the wrong thing in the path it was written for** — line 78. It checks the directory, not the file. In the non-root path, where no `chown` runs, a `filadex.db` owned by a previous uid passes the check and then produces the raw `SQLITE_READONLY` the check was meant to pre-empt. `[ ! -e "$DB_FILE" ] || [ -w "$DB_FILE" ]` would catch it.

## Minor

- The generic image entrypoint now hard-codes eleven lines of DSM-specific advice, a third copy alongside `docs/deployment-synology.md` and the compose comment, and re-parses the URL the block above already parsed. A Debian or Unraid host with `--user 1001` gets told to open File Station and set `PUID=1026`; following that chowns their data to a foreign uid on the next root start. One platform-neutral line naming the path and effective uid/gid, with a link to the Synology guide, drifts less. `[ -w "$DB_DIR" ]` also leaves no marker file behind if the container is killed mid-check.
- Please keep the sentence about snapshots being `0600` and unreadable by non-admin DSM users in the Permissions Note — `backups.ts` still implements it — and drop the "or Everyone" advice.

What is fine: the configurable ids, the DSM defaults in the compose template, the startup log line, and the mode change on the script.

Happy to push a corrected entrypoint against this branch if that is easier than a round trip.

Azumi0:
Addressed the review feedback in commit `1bfeaf7`:

1. **Backup permissions preserved**: Removed recursive `chmod -R 775` over data paths. Retained `chown -R "$PUID:$PGID"` and applied non-recursive `chmod u+rwX,g+rwX` on directory paths only (`${p}`). Snapshots retain `0600` and backup directories retain `0700`. Deduplicated paths in `DATA_PATHS`.
2. **SQLite URL normalization & reuse**: Normalized `DB_FILE` once at the entrypoint top, stripping `file:`, `sqlite:`, query parameters, and normalizing leading slashes. Only derived `DB_DIR` for absolute paths (`/*`), preventing relative (`./dev.db`) or `:memory:` URLs from targeting `/app`. Reused `DB_DIR` across both the root setup block and the preflight check.
3. **PUID/PGID validation & loop prevention**: Added numeric validation for `PUID` and `PGID` using `case ... in ''|*[!0-9]*)`. Explicitly rejected `PUID=0` with an explanatory error, and guarded `su-exec` with `[ "$PUID" != 0 ]`.
4. **Fresh volume non-root support**: Added `mkdir -p "${DB_DIR}" 2>/dev/null || true` before running the writability check.
5. **Preflight file & directory checks without marker files**: Replaced `touch` with shell `-w` checks (`[ -w "$DB_DIR" ]` and `[ ! -e "$DB_FILE" ] || [ -w "$DB_FILE" ]`).
6. **Platform-neutral guidance**: Replaced the DSM-specific error banner with a platform-neutral message identifying the unwritable path, effective UID/GID, and linking to `docs/deployment-synology.md`.
7. **Documentation restored**: Restored the statement in `docs/deployment-synology.md` noting the `0600` backup snapshot guarantee, removed `chmod -R 775` mentions, and dropped the "Everyone" permission recommendation.

the-luap (commented):
Reviewed `1bfeaf7`. All seven points are addressed, and I verified them against a real build rather than the diff alone. Fine to merge from my side.

What I ran:

- **URL parsing** through `sh` with five inputs: `file:/data/filadex.db`, `sqlite:/data/filadex.db`, `file:///data/x/filadex.db?mode=rwc`, `file:./dev.db`, `file::memory:`. The first three resolve to the right file and directory; the last two leave `DB_DIR` empty, so nothing touches `/app` any more.
- **Container with `PUID=1026 PGID=100` and a `sqlite:` URL** on a bind mount: boots, migrates, process is uid 1026 at pid 1, healthcheck reports healthy. Changed the admin password, created a backup: `-rw------- 1026 100`. Restarted the container: still `0600`. So the snapshot guarantee from #29 survives a restart now.
- **`PUID=0`** exits with the new message instead of looping. **`PUID=paul`** exits with the validation message instead of a busybox error.
- **Non-root path on a fresh volume** (`--user 1026:100`, `DATABASE_URL=file:/data/db/filadex.db`, empty `/data`): the entrypoint creates `/data/db` and the migrator runs. On `44eef03` this was the ENOENT case.

One optional tightening, not blocking: `chmod u+rwX,g+rwX` leaves `/data` and `/data/backups` at `775`. With the Synology default `PGID=100` — the `users` group, every DSM account — group write on the backups directory means any DSM user with share access can delete snapshots by browsing the share, even though they cannot read them. The app only needs the owner bits; `g+rX` without `w` would keep the directory listable for the group and nothing more. Your call whether Synology file management needs the group write.

One note for anyone verifying on a Mac: Docker Desktop's bind mounts report file ownership inconsistently inside the container (the database showed `0 0` while the directory showed `1026 100`), so judge by whether writes succeed, not by `ls -ln`.

Azumi0:
Let me know if there will be anything I need to address that would be blocking this merge, this should be merged before #32

the-luap:
Nothing blocking from my side — `1bfeaf7` resolved everything I raised, and I verified it on a real build (details above). The group-write note is optional. Agreed on the order: this before #32, which then needs a rebase for the shared `DATA_PATHS` line.

PR #31, opened 2026-09-08
Render LanguageSelector on change-password and allow language updates

Azumi0:
## Description

The `/change-password` page did not render a language selector, leaving users with no way to switch UI language during mandatory password change. In addition, the authentication middleware (`authenticate` in `server/auth.ts`) refused `POST /api/users/language` with a `403 Forbidden` (`PASSWORD_CHANGE_REQUIRED`) for accounts with `forceChangePassword: true`.

This caused two problems:
1. When a user selected a language on the login screen and then logged in with an account requiring a password change (e.g. initial `admin` setup), `LanguageProvider` attempted to persist the pending language to the account, which was rejected with 403 and produced an error toast. On page reload, the server stamped the default `en` from the unchanged database record and reverted the UI back to English.
2. Once on `/change-password`, there was no `AuthLanguageSelector` rendered on the page, preventing the user from changing the language directly.

This change:
- Renders `AuthLanguageSelector` inside `CardHeader` (with `relative` and `pr-16` for proper spacing) on `client/src/pages/change-password.tsx`, matching the layout of the other authentication views (`login`, `register`, `forgot-password`, `reset-password`, `verify-email`).
- Adds `"/api/users/language"` to `ALLOWED_WHILE_PASSWORD_CHANGE_PENDING` in `server/auth.ts`, allowing language preference to be saved and persisted while password change is pending, while keeping all other unauthorized routes blocked.

## Type of change

## How Has This Been Tested?

- [x] Component SSR integration test in `tests/components/auth-pages-language-selector.test.tsx` asserting that `ChangePasswordPage` renders `LanguageSelector` with active language label and translation key.
- [x] Route integration test in `tests/routes/users.test.ts` verifying that `POST /api/users/language` succeeds (200) and updates the account's language while password change is pending, and that other endpoints (e.g. `POST /api/users/units`) continue to be refused with 403 `PASSWORD_CHANGE_REQUIRED`.
- [x] Local verification suite:
  - `npm run check` (Postgres TypeScript): clean
  - `npm run check:sqlite` (SQLite TypeScript): clean
  - `npm run lint`: clean (0 errors, 0 warnings)
  - `npx vitest run tests/components/auth-pages-language-selector.test.tsx`: 6 passed
  - `TEST_DIALECT=sqlite npx vitest run tests/components/ tests/routes/auth.test.ts`: 83 passed

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,8 +158,9 @@ LOG_LEVEL=INFO                # Logging level (DEBUG, INFO, WARN, ERROR)
 
 # Authentication
 # The first admin is created as admin/admin whenever no admin account exists,
-# and every route except changing the password is refused until it has been
-# changed. There is no variable that sets it.
+# and every route except changing the password and setting a language
+# preference is refused until it has been changed. There is no variable that
+# sets it.
 JWT_SECRET=your_secret_key    # Secret key for JWT token generation
 
 # The address users open Filadex at. Verification and password-reset emails
--- a/client/src/i18n/LanguageProvider.tsx
+++ b/client/src/i18n/LanguageProvider.tsx
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react';
 import { useLocation } from 'wouter';
 import { LanguageContext, Language, getTranslation, interpolate } from './index';
 import { getInitialClientLanguage, isSupportedLanguage } from './resolve-language';
-import { useMutation, useQuery } from '@tanstack/react-query';
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import { apiRequest } from '@/lib/queryClient';
 import { useAuth } from '@/lib/auth';
 import { useToast } from '@/hooks/use-toast';
@@ -63,6 +63,7 @@ function writePendingAccountLanguage(language: Language | null): void {
 
 export const LanguageProvider: React.FC<LanguageProviderProps> = ({ children }) => {
   const [language, setLanguageState] = useState<Language>(getInitialClientLanguage);
+  const queryClient = useQueryClient();
   const { toast } = useToast();
   const [location] = useLocation();
   const { isPublicRoute } = useAuth();
@@ -107,6 +108,12 @@ export const LanguageProvider: React.FC<LanguageProviderProps> = ({ children }) 
         method: 'POST',
         body: JSON.stringify({ language: newLanguage })
       });
+    },
+    onSuccess: (_data, newLanguage) => {
+      queryClient.setQueryData(['/api/auth/me'], (old: any) =>
+        old ? { ...old, language: newLanguage } : old
+      );
+      queryClient.invalidateQueries({ queryKey: ['/api/auth/me'] });
     },
     onError: (error) => {
       console.error('Error updating language preference:', error);
--- a/client/src/pages/change-password.tsx
+++ b/client/src/pages/change-password.tsx
@@ -11,6 +11,7 @@ import { Input } from "@/components/ui/input";
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card";
 import { Form, FormControl, FormField, FormItem, FormLabel, FormMessage } from "@/components/ui/form";
 import { useToast } from "@/hooks/use-toast";
+import { AuthLanguageSelector } from "@/components/language-selector";
 
 // Create a function to generate the schema with translations
 const createChangePasswordSchema = (t: (key: string) => string) => z.object({
@@ -82,7 +83,8 @@ export default function ChangePasswordPage() {
   return (
     <div className="flex items-center justify-center min-h-screen bg-background">
       <Card className="w-[400px]">
-        <CardHeader className="bg-primary dark:bg-primary text-white rounded-t-lg">
+        <CardHeader className="relative bg-primary dark:bg-primary text-white rounded-t-lg pr-16">
+          <AuthLanguageSelector />
           <CardTitle>{t('auth.changePassword')}</CardTitle>
           <CardDescription className="text-white/80">{t('auth.changePasswordDescription')}</CardDescription>
         </CardHeader>
--- a/client/src/pages/change-password.tsx
+++ b/client/src/pages/change-password.tsx
@@ -81,9 +81,9 @@ export default function ChangePasswordPage() {
   }
 
   return (
-    <div className="flex items-center justify-center min-h-screen bg-background">
-      <Card className="w-[400px]">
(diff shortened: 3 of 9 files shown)


the-luap (commented):
The change is right in substance: the pending-language flush from the login screen hitting the 403 was my omission when I wrote the allowlist, and a language preference is a harmless thing to let a password-change-pending account write. The selector on the page matches the login header's pattern. Two things should change before merge, and a few smaller ones are worth a minute.

## Before merge

**1. The forced-change test no longer proves admission** — `tests/routes/users.test.ts:106`

The closing assertion after the password change still probes `POST /api/users/language` with the fresh cookie — but this PR makes that route succeed *while the flag is still set*, so it now passes whether or not the account was admitted. If `change-password` ever stopped clearing `forceChangePassword`, this test would stay green. Probe a route that is refused before the change, e.g. `POST /api/users/units` as line 82 already does. (Line 91's bare `GET /api/auth/me` and the inline `me.body.language` read are also what the file's own `currentUser(cookie)` helper at line 47 is for, and the rewritten comment promises a theme check the test does not make.)

**2. The saved language is overwritten on the way to the home page** — `client/src/i18n/LanguageProvider.tsx:104`

`persistAccountLanguage` neither invalidates nor updates the `['/api/auth/me']` query. On `/change-password` the provider has cached `{ language: "en" }`; pick Polish (POST 200, cache untouched); submit; `navigate("/")` remounts `LanguageProvider` under `AuthProvider`'s placeholder; the query hands back the cached `en` synchronously, the `[account]` effect applies it, and `<html lang>` and the cookie flip back to English until the background refetch returns `pl`. This path is newly reachable because of this PR. `use-units.tsx:38` and `settings-notifications.tsx:61` already invalidate that key after their writes; an `onSuccess` doing the same here (or `setQueryData`) closes it.

## Worth a minute

- **Docs still state the old contract.** README:161, `docs/deployment-synology.md:45` and the comment on `bootstrapAdmin` in `tests/helpers/app.ts` say every request except the change itself is refused. That sentence should now mention the language preference, or an operator checking the bootstrap will file the 200 as a bug.
- **The allowlist is path-only for this entry** — `server/auth.ts:91`. `Set.has(req.path)` admits every method, unlike the deliberately method-scoped `GET /api/theme` branch just below. Fine today; a `"${method} ${path}"`-keyed set would make the rule explicit and drop the special case. Related: a theme toggle next to the selector would `POST /api/theme` and fail silently, since the redirect helper skips when already on `/change-password` and the theme mutation has no `onError`.
- **Header layout.** `pr-16` reserves 64px for a control that takes roughly 78px with its offset, and this is now the only auth header that compensates with padding rather than using the siblings' `text-center` header string. The card also keeps its fixed `w-[400px]` without the `p-4` wrapper the other auth pages have, which is pre-existing but is what clips the new control on a 360px phone.
- **Coverage.** Bug 1 in the description — pick a language on `/login`, log in as a force-change account, watch it revert — is effect ordering in a real browser, which CLAUDE.md §5 says to pin with Playwright. `tests/e2e/login-language.spec.ts` is the template and `bob` is already seeded with `forceChangePassword: true`, so it is a short spec. The `renderToString` test proves the selector renders; it cannot see the flush.

One pre-existing note, not this PR's: the pending choice is cleared from `sessionStorage` before the POST resolves, so any failure of that request other than the 403 (limiter, network) still loses it.

Azumi0:
Pushed commit 17cae91 addressing the review findings:

### Before merge
1. **Admission proof in `tests/routes/users.test.ts`**: The closing assertion now probes `POST /api/users/units` with the fresh cookie to prove admission (which is refused with 403 prior to password change). Also cleaned up the inline `GET /api/auth/me` calls in favour of the file's `currentUser` helper, and aligned the comment with what is checked.
2. **Query cache synchronization in `LanguageProvider.tsx`**: `persistAccountLanguage` now updates `['/api/auth/me']` synchronously in `onSuccess` via `queryClient.setQueryData` and marks the query stale with `invalidateQueries`, preventing stale cached data from flipping `<html lang>` and cookies on navigation to `/`.

### Worth a minute
- **Contract documentation**: Updated `README.md`, `docs/deployment-synology.md`, and `bootstrapAdmin` in `tests/helpers/app.ts` to document that updating language preference is permitted while password change is pending.
- **Method-scoped allowlist**: Refactored `ALLOWED_WHILE_PASSWORD_CHANGE_PENDING` in `server/auth.ts` to a `"${method} ${path}"`-keyed set (`POST /api/auth/change-password`, `GET /api/auth/me`, `POST /api/auth/logout`, `POST /api/users/language`, `GET /api/theme`).
- **Header layout**: Added `p-4` to the container on `/change-password`, switched the card to `w-full max-w-[400px]` to avoid clipping on 360px viewports, and aligned `CardHeader` with `text-center` consistent with other auth headers.
- **Coverage**: Added Playwright E2E spec in `tests/e2e/change-password-language.spec.ts` pinning the login language flush for an account with `forceChangePassword: true` (`bob`) and verifying it persists across page reload on `/change-password`. Cleans up `bob`'s language back to `de` in `finally` without modifying passwords or fixture state.

All suites pass locally: `npm run check`, `npm run check:sqlite`, `npm run lint`, `vitest run` (453 passed), and `playwright test` (7 passed).

the-luap (commented):
Reviewed `17cae91`. Both before-merge items and all four smaller ones are addressed. Fine to merge from my side.

Verified on the branch head:

- `npm run check`, `npm run lint` clean; `tests/routes/users.test.ts` and `tests/components` pass under SQLite (one run hit the register → verify → login flake once; two reruns clean).
- `npm run build && npm run test:e2e` — 7 passed, including the new `change-password-language.spec.ts`.
- **The new spec pins the fix.** With `POST /api/users/language` removed from the allowlist and the bundle rebuilt, it fails at the stored-language assertion with `Expected: "pl", Received: "de"`; with the head it passes. So it is the allowlist entry that makes it work, not the render.
- The admission assertion now probes `POST /api/users/units` with the fresh cookie and checks `forceChangePassword` is cleared, which is what proves the account got past the gate.
- `setQueryData` plus `invalidateQueries` in `onSuccess` covers the remount on the way to `/`.
- The method-keyed allowlist reads clearly and drops the special case; the docs and the helper comment now state the actual contract.

Small: the spec's `finally` restores `bob` through the same `POST /api/users/language` this PR allowlists, which is neat, and it never touches the password — so the seeded fixture stays intact for the other specs.

Azumi0:
Let me know if there will be anything I need to address that would be blocking this merge

the-luap:
Nothing blocking — `17cae91` addressed all of it and the new spec pins the fix (verified above). Ready to merge from the review side.

PR #32, opened 2026-09-08
Add Open Filament Database (OFD) catalog support and barcode/GTIN scanning

Azumi0:
## Overview

Integrates the [Open Filament Database](https://openfilamentdatabase.org) (OFD) community catalog alongside SpoolmanDB on the `add-filament` modal, introduces barcode/GTIN scanning and automatic lookup, transitions community catalog caching to a lightweight server-side in-memory index backed by disk persistence (ADR 0006), and adds home collection barcode scanning and filtering.

---

## Architectural Changes (ADR 0006)

- **In-Memory Catalog Index with Atomic Disk Persistence** (`server/services/community-catalog.ts`):
  - Retires the relational `community_filament_cache` table to prevent database bloat (>35k community records) and heavy TRUNCATE/INSERT churn on SQLite homelab instances.
  - Loads from gzip-compressed OFD dataset (`all.json.gz`) and SpoolmanDB vendor files. Searches take ~1.3ms in memory across both providers.
  - Persists atomic disk snapshots (`ofd.json` and `spoolmandb.json`) using temporary files and rename swaps with cleanup guards. Default cache directory configurable via `CATALOG_CACHE_DIR`, defaulting to `/data/cache/catalogs` in production Docker.
  - Daily background sync scheduler (`server/catalog-scheduler.ts`) checks every 24 hours with mutex locking, provider isolation in `sync("all")`, and a 30-minute failure cooldown to prevent API rate-limit retry storms.
  - Resilience safeguards: 0-item sync guard for OFD and >50% failure threshold for SpoolmanDB downloads prevent wiping valid disk caches during upstream outages.
- **Dual-Dialect Database Parity**:
  - Adds `barcode` text column to `filaments` table (`0008_mighty_mephistopheles.sql` / `0002_rich_nehzno.sql`).
  - Drops obsolete `community_filament_cache` table (`0009_round_radioactive_man.sql` / `0003_eager_william_stryker.sql`).
  - Strict parity verified: 0 migration drift across PostgreSQL and SQLite.

---

## Features & UX Improvements

1. **Add Filament Screen**:
   - Tab toggle between **Open Filament Database** and **SpoolmanDB** above the community search input, persisting user preference in `localStorage` (defaulting to OFD).
   - Dedicated `Barcode` input field in form.
   - Integrated camera scanner routes scanned barcodes: Filadex QR -> Local Collection -> OFD GTIN lookup. If not found in any source, a toast alert cites all searched sources (Filadex QR, Bambu Lab, Open Filament Database) and retains the scanned code in the barcode field.
   - Pre-fills material, manufacturer, name, colors, nozzle/bed temperatures (prioritizing OrcaSlicer, Bambu Studio, PrusaSlicer, and Cura settings), density, diameter, spool type (refill vs spooled), weight, and barcode.
2. **Home Collection Search & Scanner**:
   - Filter sidebar includes a dedicated barcode scan button next to the search input.
   - Zero-padded GTIN normalization (`client/src/lib/filament-search.ts`) enables matching 14-digit GTIN barcodes with 13-digit EAN records and vice versa, while guarding against all-zero edge cases.
3. **Settings Community Filaments**:
   - Status cards for both OFD and SpoolmanDB displaying item counts, last update timestamps, and options for provider-specific or global refresh.
4. **Internationalization**:
   - Complete translations across English, German, and Polish.

---

## Verification

- `npm run check` -> Clean (0 TypeScript errors for PostgreSQL)
- `npm run check:sqlite` -> Clean (0 TypeScript errors for SQLite)
- `npm run lint` -> Clean (0 ESLint errors/warnings)
- `npm run db:check-drift` -> Clean (0 migration drift across PostgreSQL and SQLite)
- `npm test` -> 34/34 test suites passing (476 passed, 9 skipped, 0 failed)


the-luap (commented):
This is a big, well-structured piece of work — the in-memory catalog with atomic disk snapshots (ADR 0006) is the right replacement for a 35k-row cache table, the routes are all authenticated, the search limit is capped, the GitHub token goes only to `api.github.com`, and CI is green. I would still not merge it yet: the headline feature does not work for the common case, and three things the previous sync had were lost in the rewrite. The first three items below I reproduced myself; the rest were confirmed by a second pass against the branch and the live dataset.

## Blocking

**1. Scanning a numeric barcode in the Add modal does nothing** — `client/src/components/filament-modal.tsx:525`

`handleQRCodeScanned` starts with `JSON.parse(decodedText)`. A GTIN such as `6975337031901` *is* valid JSON — it parses to a number — so `applyScannedFilamentData(6975337031901)` bails on `!data.name`, and the `return` inside the `try` exits before the GTIN lookup, the barcode field write and either toast. The scanner closes and the form is unchanged. Every EAN/UPC whose first digit is 1–9 takes this path; only zero-padded codes fail the parse and fall through. Check `typeof parsed === "object"` before treating it as scan data.

**2. OFD temperatures are never read** — `server/services/community-catalog.ts:243-262`

The parser looks for `slicer_settings[orca|bambu_studio|prusa_slicer|cura].extruder_temp`. The live `all.json.gz` (today's build, 2089 filaments) keys slicers as `prusaslicer` etc., the entries carry `profile_name` only, and temperatures live at top level as `min/max_print_temperature` and `min/max_bed_temperature` — present on 1585 filaments. `extruder_temp` appears in 0 slicer entries. So `extruderTemp` is null for every OFD item and the prefill described in the PR never happens from OFD.

**3. Barcodes are lost on every export/import round-trip** — `server/routes/filaments.ts`

`barcode` is wired into single create (303) and PATCH (384) only. The CSV export columns, the CSV import field list and the JSON import `insertData` were not extended, so a reinstall from an export produces spools with `barcode = NULL`, and the home-page scanner then never matches an existing spool.

**4. The sync lost the validation and the size cap #28 added** — `community-catalog.ts:144, 266, 331, 484`

The deleted `spoolmandb-sync.ts` validated each vendor file with zod and skipped malformed ones, and capped responses at 8 MiB. The replacement calls `col.hex.startsWith`, `variant.color_hex.startsWith` and `item.gtin.trim()` after truthiness checks only, and the per-file `try` wraps fetch and `JSON.parse` but not the parser — so one wrongly-typed upstream field throws out of the whole source sync, `POST /refresh` answers 500 and the scheduler enters its 30-minute cooldown on every attempt. A numeric `gtin` hits the same path in `rebuildIndexes`, which also runs from `loadFromDisk`, so a bad cached value breaks startup load. The OFD path reads the full body, `gunzipSync` with no `maxOutputLength`, then `JSON.parse`, all synchronously on the event loop; the normal payload is already 2.9 MB gzip → 11 MB JSON, and an oversized or malicious response has no ceiling. Please bring back a schema per record (skip the record, not the source) and a byte cap, and give `gunzipSync` a `maxOutputLength`.

## Should fix

- **Every failure reads as "not recognized"** — `home.tsx:405`, `filament-modal.tsx:541`. `apiRequest` throws on 404, so `if (result)` is dead and the bare `catch {}` turns a 401 (expired session), a 500 or a network error into the destructive "not recognized" toast and an empty Add modal. The thrown error carries `status`; branch on 404.
- **SpoolmanDB results force 1 kg / spooled** — `community-catalog.ts:344`. The parser emits `weightGrams: 1000, spoolRefill: false` for a source that has neither, so the client's null guards never fire and picking a SpoolmanDB result overwrites a user's 250 g / refill entry. `CommunityCatalogItem` already types both as nullable; emit null for unknown.
- **Filadex's own labels are not handled by the new scan paths** — `home.tsx:282`. Labels encode `${origin}/?openFilament=<id>`; neither scan handler has a URL branch, so scanning a printed label ends in "not recognized" with the URL in the barcode field instead of the existing deep-link.
- **Bambu 1D codes never persist their barcode** — `home.tsx:320`. The scanner converts `BBL-…` into an object without the raw code, so the spool is created with `barcode: ""` and a rescan can only match by exact name.
- **Shared GTINs resolve to an arbitrary variant** — `community-catalog.ts:148`. `nextGtinMap.set(norm, item)` lets the last record win. In today's OFD data 88 GTINs are shared by 2–5 size records, differing in colour (49), spool vs refill (30) or brand (18). A scan of one of those silently prefills the other variant. Keep all candidates and let the client choose, or prefer the record whose size matches.

## Coordination

- **Conflicts with #30**: both touch the same `DATA_PATHS=` line in `docker-entrypoint.sh`. Whichever merges second needs a rebase; the resolution is the union.
- There is uncommitted local work on `dev` in five of these files (`settings-community-filaments.tsx`, `community-filaments.ts`, `storage.ts`, `de.ts`, `en.ts`) that this PR rewrites — noted here so it is not a surprise at merge time.

## What is fine

Route auth and the admin gate on refresh, the search limit cap, fixed upstream hosts, the token scoping, the `.tmp` + rename snapshot write, the zero-item and failure-threshold guards, the migration pair with a clean drift check, and the translations for all three languages.

the-luap:
#30 and #31 are merged, so this now conflicts on the `DATA_PATHS=` line in `docker-entrypoint.sh`. The resolution is the union of both additions: keep #30's PUID/PGID handling and add `${CATALOG_CACHE_DIR:-/data/cache/catalogs}` to the list. The review findings above still stand.

Azumi0:
Addressed all blocking and suggested review findings, rebased on `main`, and validated with the full test suite and browser e2e suite.

### Summary of Changes

1. **Numeric barcode handling in Add modal** (`client/src/components/filament-modal.tsx`)
   - Guarded `JSON.parse` output to require a non-null object with either `name` or `material`. Numeric or non-filament JSON payloads fall through to barcode/GTIN lookup as intended.

2. **OFD temperature parsing** (`server/services/community-catalog.ts`)
   - Switched from looking up `slicer_settings.*.extruder_temp` to parsing the top-level `min_print_temperature`, `max_print_temperature`, `min_bed_temperature`, and `max_bed_temperature`.
   - Midpoint printing temp is assigned to `extruderTemp`, and `bedTemp` is extracted from max bed temperature (or midpoint when `min > 30°C`).

3. **Barcode export/import round-tripping** (`server/routes/filaments.ts`)
   - Added `barcode` to CSV export headers and data rows.
   - Added `barcode` to CSV import column map and row insertion.
   - Added `barcode` to JSON import mapping.

4. **Sync validation & response caps** (`server/services/community-catalog.ts`)
   - Reintroduced per-record Zod validation (`spoolmanDbFilamentSchema`, `spoolmanDbVendorFileSchema`, `ofdBrandSchema`, `ofdFilamentSchema`, `ofdVariantSchema`, `ofdSizeSchema`), skipping malformed records instead of aborting the sync.
   - Added response byte caps: 8 MiB for SpoolmanDB vendor files, 15 MiB for OFD gzip payloads, and `maxOutputLength: 50 * 1024 * 1024` (50 MiB) on `gunzipSync`.
   - Ensured GTIN indexing and cache hydration safely sanitize and coerce string keys, preventing non-string inputs from throwing.

5. **Scan error differentiation** (`client/src/pages/home.tsx`, `client/src/components/filament-modal.tsx`)
   - Branched error handling to distinguish 404 ("not recognized", opening the add modal / manual entry) from network, 401, or 500 errors, which now surface an explicit lookup error toast.

6. **SpoolmanDB default weight and refill attributes** (`server/services/community-catalog.ts`)
   - Unknown `weightGrams` and `spoolRefill` from SpoolmanDB are emitted as `null` rather than default 1000g / spooled values, preserving client-side entries and null guards.

7. **Filadex QR deep-link labels** (`client/src/pages/home.tsx`, `client/src/components/filament-modal.tsx`)
   - Added regex check for `[?&]openFilament=(\d+)` in both scan handlers to properly open existing spool details instead of falling into GTIN lookup.

8. **Bambu 1D barcode persistence** (`client/src/components/qr-scanner.tsx`)
   - Preserved raw scanned code as `barcode: barcode` in `processBambuLabBarcode` (and `barcode: qrCode` in `processBambuLabQRCode`).

9. **Shared GTIN variant candidate picker** (`server/services/community-catalog.ts`, `server/routes/community-filaments.ts`, `client/src/pages/home.tsx`, `client/src/components/filament-modal.tsx`)
   - `gtinMap` now maps GTINs to lists of candidates (`CommunityCatalogItem[]`).
   - `/api/community-filaments/gtin/:code` returns `{ ...primary, candidates }`.
   - When multiple candidates match a scanned GTIN, a selection dialog prompts the user to pick the exact variant (brand, material, color, spool/refill).

10. **Rebase & docker-entrypoint coordination**
    - Rebased onto `main` (`b8a454b`).
    - Merged #30's PUID/PGID handling with `${CATALOG_CACHE_DIR:-/data/cache/catalogs}` in `DATA_PATHS`.

### Verification

- `npm run check`: 0 errors
- `npm run check:sqlite`: 0 errors
- `npm run lint`: 0 errors, 0 warnings
- `npm run db:check-drift`: Clean (PostgreSQL & SQLite)
- `npm test`: 34 suites passed, 482 passed, 9 skipped, 0 failed
- `npm run build && npm run test:e2e`: 14 browser tests passed

the-luap (commented):
Reviewed `bd24fe9` and `39697d2` on the branch head. All four blocking items and the five should-fix items are addressed, the rebase resolved the entrypoint line the way #30 needs it, and I verified the substantive ones rather than the diff alone. Ready to merge from the review side.

What I checked:

- **OFD temperatures, against the live dataset.** Ran the branch's `CommunityCatalogService.sync("ofd")` against today's `all.json.gz` (22354 items) and searched for PLA: 376 of 400 results now carry an extruder temperature and a bed temperature, where the previous head produced none. The midpoint-of-range rule reads sensibly (`123-3D Filament PLA` → 205 °C / 50 °C).
- **Shared GTINs.** `lookupGtinCandidates("6975337032878")` returns both Bambu ABS records, spool and refill, so the picker has something to pick from instead of whichever came last in the file.
- **Numeric barcodes.** The JSON branch now requires an object with `name` or `material`; a bare number falls through to the GTIN lookup, and a Filadex label URL is handled before either. Verified by reading — there is no automated path through a camera scan, so this one stays a code-read.
- **Barcode round-trip.** Present in the CSV export header and row, the CSV import column map and insert, and the JSON import.
- **Sync safety.** Per-record zod schemas with `safeParse` that skip the record, an 8 MiB cap per SpoolmanDB file, a 15 MiB cap on the OFD body and `maxOutputLength` on `gunzipSync`.
- **Error handling** branches on `err.status !== 404` in both scan paths; SpoolmanDB emits `null` for unknown weight and refill; the Bambu 1D path keeps the raw code.
- `npm run check`, `npm run lint`, `npm run db:check-drift` clean; `TEST_DIALECT=sqlite npx vitest run` — 490 passed, 1 skipped; `npm run build && npm run test:e2e` — 14 passed.

Not run locally: the Postgres suite, because Docker on my machine stopped answering mid-run. CI's Postgres job on this head is green, which covers it.

One thing outside this PR's control: there is uncommitted local work on `dev` in five of the files this PR rewrites, so the maintainer will need to reconcile that at merge time. Not a blocker for the PR itself.

Azumi0:
Let me know if there will be anything I need to address that would be blocking this merge

the-luap:
Nothing blocking — `39697d2` resolved all four blocking items and the five smaller ones, and I verified the substantive ones against the live OFD dataset (details in my review above). Ready to merge.

One sequencing note: #33 currently contains this branch in full, so merging this one first will shrink #33 to just the registration work.

PR #33, opened 2026-09-09
feat: add registration policy with admin toggle and E2E coverage

Azumi0:
## Overview

Adds a runtime registration policy with an administrative toggle, allowing administrators to disable public self-registration while preserving manual administrator provisioning (ADR 0006).

When self-registration is disabled:
- POST /api/auth/register returns 403 Forbidden (REGISTRATION_DISABLED).
- The login page hides the Create account link.
- Direct visits to /register display an informational notice explaining that registration is disabled and provide a Back to login action.
- Administrators can continue to provision user accounts manually in the User Management interface.

---

### 1. Database & Schema
- Singleton system_settings table (registration_enabled boolean default true).
- PostgreSQL migration (0010_jazzy_korvac.sql) and SQLite migration (0004_misty_guardsmen.sql).
- Drizzle schemas and Zod validation schemas (shared/schema.ts).
- Dual-dialect storage support in server/storage.ts.

### 2. API Endpoints
- GET /api/system/public-settings — unauthenticated endpoint returning public instance flags ({ registrationEnabled }).
- GET /api/settings/system — admin-only endpoint returning instance settings.
- PUT /api/settings/system — admin-only endpoint to update system settings with Zod validation.
- Guard in POST /api/auth/register checking storage.getSystemSettings().registrationEnabled.

### 3. UI & UX Improvements
- **Header**: Updated List Management menuitem to General Settings with the SlidersHorizontal icon.
- **User Management Modal**:
  - Overhauled user list to a responsive card layout featuring user avatars, username, role badges (Admin / User), last login timestamp, and edit/delete actions.
  - Added registration policy switch card (Allow new user registration) with real-time toggle mutation and toast notifications.
- **Login & Register Views**:
  - LoginPage queries /api/system/public-settings and conditionally hides the registration link when disabled.
  - RegisterPage renders an informational card when disabled with Filadex branding and a Back to login navigation button.
- **Internationalization**:
  - Complete translations added in English (en.ts), German (de.ts), and Polish (pl.ts).

### 4. Testing & Verification
- Unit & Route tests: tests/routes/system-settings.test.ts (GET / PUT routes, validation, permissions).
- Component SSR tests: tests/components/registration-disabled-ui.test.tsx, tests/components/user-management-modal.test.tsx, tests/i18n/general-settings-keys.test.ts.
- Full Playwright E2E coverage in tests/e2e/registration-policy-and-user-management.spec.ts:
  - Verified Header Settings and Tools role visibility.
  - Verified UserManagementModal switch, card list, role badges, and edit transitions.
  - Verified end-to-end disabling and re-enabling flow across /login and /register.

---

## Verification

- npm run check -> Clean (0 TypeScript errors for PostgreSQL)
- npm run check:sqlite -> Clean (0 TypeScript errors for SQLite)
- npm run lint -> Clean (0 ESLint errors/warnings)
- npm test -> 38/38 suites passing (498 passed, 9 skipped, 0 failed)
- npx playwright test tests/e2e/registration-policy-and-user-management.spec.ts -> 4/4 passed

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -83,3 +83,20 @@ _Avoid_: External database, Global catalog
 A machine-readable code (GTIN, EAN, UPC, or vendor barcode) identifying a
 spool or its retail packaging, recorded on the spool.
 _Avoid_: QR code (when referring to 1D barcodes), GTIN (when referring to non-standard vendor barcodes)
+
+### Accounts and access
+
+**Self-Registration**:
+The public flow where an unauthenticated visitor creates their own user
+account, requiring email verification before login.
+_Avoid_: Sign up, Public registration, Guest creation
+
+**Admin Provisioning**:
+The administrative flow where an administrator directly creates an account,
+optionally assigning administrative privileges and setting a temporary password.
+_Avoid_: Manual creation, User invite
+
+**Registration Policy**:
+The instance-wide setting determining whether Self-Registration is permitted.
+_Avoid_: Registration toggle, Sign-up switch
+
--- a/client/src/components/header.tsx
+++ b/client/src/components/header.tsx
@@ -11,7 +11,7 @@ import {
   User,
   Wrench,
   ChevronDown,
-  List,
+  SlidersHorizontal,
   Gauge,
   ClipboardList
 } from "lucide-react";
@@ -94,8 +94,8 @@ export function Header({ onAddFilament }: HeaderProps) {
                 setSettingsDialogTab(undefined);
                 setSettingsDialogOpen(true);
               }}>
-                <List className="mr-2 h-4 w-4" />
-                {t('settings.listManagement')}
+                <SlidersHorizontal className="mr-2 h-4 w-4" />
+                {t('settings.generalSettings')}
               </DropdownMenuItem>
               <DropdownMenuItem onClick={() => {
                 setSettingsDialogTab('units');
--- a/client/src/components/user-management-modal.tsx
+++ b/client/src/components/user-management-modal.tsx
@@ -10,11 +10,13 @@ import { Dialog, DialogContent, DialogHeader, DialogTitle, DialogDescription } f
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Checkbox } from "@/components/ui/checkbox";
+import { Switch } from "@/components/ui/switch";
+import { Badge } from "@/components/ui/badge";
+import { Label } from "@/components/ui/label";
 import { Form, FormControl, FormField, FormItem, FormLabel, FormMessage } from "@/components/ui/form";
 import { useToast } from "@/hooks/use-toast";
-import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from "@/components/ui/table";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
-import { Trash2, Edit, UserPlus } from "lucide-react";
+import { Trash2, Edit, UserPlus, User as UserIcon } from "lucide-react";
 
 // Create a function to generate the schema with translations.
 // The username rules mirror usernameSchema in shared/schema.ts, which both
@@ -62,6 +64,37 @@ export function UserManagementModal({ open, onOpenChange }: { open: boolean; onO
     queryKey: ["users"],
     queryFn: () => apiRequest("/api/users"),
     enabled: open,
+  });
+
+  const { data: systemSettings, isLoading: isLoadingSettings } = useQuery<{ registrationEnabled: boolean }>({
+    queryKey: ["/api/settings/system"],
+    queryFn: () => apiRequest<{ registrationEnabled: boolean }>("/api/settings/system"),
+    enabled: open,
+  });
+
+  const toggleRegistrationMutation = useMutation({
+    mutationFn: (enabled: boolean) =>
+      apiRequest("/api/settings/system", {
+        method: "PUT",
+        body: { registrationEnabled: enabled },
+      }),
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ["/api/settings/system"] });
+      queryClient.invalidateQueries({ queryKey: ["/api/system/public-settings"] });
+      toast({
+        title: t('common.success'),
+        description: t('users.updateSuccess'),
+        variant: "success",
+      });
+    },
+    onError: (error: any) => {
+      const errorMessage = getErrorMessage(error);
+      toast({
+        title: t('common.error'),
+        description: errorMessage,
+        variant: "destructive",
+      });
+    },
   });
 
   const form = useForm<UserFormValues>({
@@ -222,7 +255,7 @@ export function UserManagementModal({ open, onOpenChange }: { open: boolean; onO
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogContent
-        className="sm:max-w-[600px]"
+        className="w-[95vw] sm:max-w-[650px] max-h-[90vh] overflow-y-auto p-4 sm:p-6"
         aria-describedby="user-management-description"
       >
         <DialogHeader>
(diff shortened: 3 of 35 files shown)


Azumi0:
This is blocked by #32 , please merge that one first, then I can rebase this one and you will be able to review it.

the-luap (commented):
The registration policy is a sensible feature and the UI/route/i18n work looks complete. Three things block it, and the first two I reproduced on the branch. Note also that this PR **contains all of #32** (its head is an ancestor), so merging this merges that too — see the coordination note at the end.

## Blocking

**1. The settings singleton races on first use, and an unauthenticated endpoint triggers it** — `server/storage.ts:660-668`

`getSystemSettings()` does a `select`, then an `insert` when the row is missing, with nothing making that atomic. I ran eight concurrent calls against an empty table on the branch:

```
concurrent: 1 ok, 7 rejected
Failed query: insert into "system_settings" ("id", "registration_enabled", "updated_at") ...
```

Seven of eight fail on the primary key. This is not a theoretical race: `GET /api/system/public-settings` is unauthenticated and the login page calls it on load, so on a fresh install two tabs, a reload during load, or a health check overlapping a visitor produces 500s — and `updateSystemSettings` calls `getSystemSettings()` first, so the admin toggle can fail the same way. Use an upsert (`onConflictDoNothing` then re-select), or seed the row in the migration the way `0001_seed_email_settings_row.sql` already does for email settings — that is the established pattern here and it removes the lazy insert entirely.

**2. `DISABLE_RATE_LIMITS` is a production kill switch added for the test suite** — `server/utils/rate-limits.ts:6`, `playwright.config.ts:88`

`skip: skipIfDisabled` is evaluated per request, so setting `DISABLE_RATE_LIMITS=true` turns off all four limiters, the login limiter included. That is the brute-force protection #27 added, and the switch is undocumented in `.env.example`, so nothing warns an operator who copies it from the Playwright config into a compose file. The browser suite does need relief here, but it should not be reachable in production: gate it on `process.env.NODE_ENV === "test"` (which `vitest.config.ts` already sets and the e2e server does not), or raise the limits for the e2e run rather than removing them.

Related, smaller: `skipSuccessfulRequests: true` on the login limiter means a successful login costs nothing against the budget. Reasonable for usability, but it is a deliberate loosening of that limiter and deserves a line of comment saying so.

**3. `DISABLE_REGISTRATION` silently does nothing after first boot** — `server/storage.ts:662`

The variable is only read when the row is created. On the branch:

```
after setting DISABLE_REGISTRATION=true on an existing install: registrationEnabled = true
```

An operator who sets it on a running install gets no effect and no message. It is also absent from `.env.example`, the README and the Synology guide. Either make it authoritative at every boot (and say that it overrides the admin toggle), or drop it and let the admin switch be the only control — the switch is the feature, and one control is easier to explain than two that disagree.

## Should fix

- **No limiter on the three new routes** — `server/routes/system-settings.ts`. `/api/system/public-settings` is unauthenticated and does a database read per call (and on a fresh install an insert). `publicReadLimiter` already exists for exactly this shape; `sensitiveActionLimiter` would suit the admin `PUT`.
- **Duplicate ADR number.** `docs/adr/0006-community-catalogs-in-memory-cache.md` (from #32) and `docs/adr/0006-system-settings-and-registration-policy.md` both claim 0006. This one should be 0007.

## Not a problem

The failing `e2e` check is infrastructure, not a test: the job died in `apt-get` during the Chromium install (`Some index files failed to download`, `Installation process exited with code: 100`). Worth a re-run rather than a fix.

## Coordination

`git merge-base --is-ancestor` confirms this branch contains #32 in full; the delta here is 32 files. I reviewed #32 separately and found it ready, so the cleanest path is to merge #32 first and let this one reduce to the registration work, rather than approving 9k lines of catalog code a second time under a different title. That also keeps the two features separable if one has to be reverted.

Local state, not this PR's problem: there is uncommitted work on `dev` in five files both PRs rewrite.

the-luap:
**Correction to my note about the `e2e` check.** I said it was an apt-get flake and only needed a re-run. That was true of the first run, which died installing Chromium before any test ran — but I re-ran the job and the tests underneath fail for real. Sorry for the misdirection.

`registration-policy-and-user-management.spec.ts:24` — "hides 'User Management' from Tools menu for non-admin users" — fails on both the attempt and the retry:

```
Locator: getByRole('button', { name: /^tools$/i })   → element(s) not found
Retry #1: getByRole('button', { name: /^settings$/i }) → element(s) not found (inside signIn)
```

**Cause: the spec signs in as `alice`, and an earlier spec leaves her account in Polish.** `login-language.spec.ts` asserts `storedLanguage(page)` becomes `"pl"` and never restores it. Playwright runs `workers: 1` against one shared database in alphabetical file order, and `login-language` (9th) runs before `registration-policy-and-user-management` (13th). So by the time this spec runs, alice's stored preference is `pl`, the header renders `Narzędzia` and `Ustawienia`, and both English-only locators miss. The retry fails one step earlier, inside `signIn`, for the same reason.

Three ways out, in the order I would prefer them:

1. **Sign in as `carol`** — seeded `role: "user"`, `forceChangePassword: false`, and no other spec touches her. `register-language.spec.ts` already picked `bob` over `alice` for exactly this reason, and the repo's own `CLAUDE.md` says to do this when a spec mutates account state.
2. Have `login-language.spec.ts` restore alice in a `finally`, the way `change-password-language.spec.ts` does for bob.
3. Match both languages in the locator, as `helpers.ts` already does with `/general settings|list management/i`. This works but hides the coupling rather than removing it.

Also visible in that run: `materials-attention.spec.ts:50` failed once and passed on retry, so it is reported flaky rather than failing. Same shared-state family, worth a look while you are in there.

The three blocking findings in my review above are unchanged.

the-luap:
#32 is merged, so this branch's catalog commits are now in `main`. Next steps as I see them:

**1. Rebase on `main`.** That drops the ~9k lines of catalog work from the diff and leaves the 32-file registration change, which is what this PR is actually about and what is left to review.

**2. The three blockers from my review**, all of which live in that remaining delta:

- `getSystemSettings()` select-then-insert races on first use — eight concurrent calls on an empty table gave one success and seven primary-key failures, and the unauthenticated `/api/system/public-settings` the login page calls is what triggers it on a fresh install. Seeding the row in the migration, the way `0001_seed_email_settings_row.sql` already does for email settings, removes the lazy insert entirely.
- `DISABLE_RATE_LIMITS` disables all four limiters at runtime, login included. Gate it on `NODE_ENV === "test"` or raise the limits for the e2e run instead.
- `DISABLE_REGISTRATION` is read only when the row is created, so on an existing install it silently does nothing. Either make it authoritative every boot or drop it and let the admin toggle be the single control.

**3. The failing e2e spec**, diagnosed in my previous comment: it signs in as `alice`, whom `login-language.spec.ts` leaves set to Polish, so the English-only `Tools` and `Settings` locators miss. Signing in as `carol` is the smallest fix.

Also worth folding in: the duplicate ADR number (this one should be 0007, since #32 took 0006), and a limiter on the three new routes.

No rush from my side — ping me when it is rebased and I will re-review the registration delta on its own.

Azumi0:
Rebased on `main` and addressed all review findings:

1. **Rebase on `main`**:
   - Rebased cleanly on top of `0efecb6`, dropping the merged catalog commits and reducing the diff strictly to the registration policy changes.

2. **`getSystemSettings()` Concurrency Race**:
   - Seeded the singleton row in both PostgreSQL (`migrations/pg/0010_jazzy_korvac.sql`) and SQLite (`migrations/sqlite/0004_misty_guardsmen.sql`) using `INSERT ... ON CONFLICT (id) DO NOTHING`.
   - Added `systemSettings` seeding to `scripts/seed.ts` for starter and demo environments.
   - Updated `server/storage.ts` fallback insertion to use `.onConflictDoNothing()` followed by a re-select, eliminating duplicate key errors on concurrent first-use calls against empty tables.
   - Added an 8-way concurrent invocation test on an empty table in `tests/routes/system-settings.test.ts`.

3. **`DISABLE_RATE_LIMITS` & Headroom Scaling**:
   - In `server/utils/rate-limits.ts`, gated `skipIfDisabled` on `process.env.NODE_ENV === "test" && process.env.DISABLE_RATE_LIMITS === "true"` so it cannot bypass limiters in production.
   - Introduced `RATE_LIMIT_MULTIPLIER` support. In `playwright.config.ts`, replaced `DISABLE_RATE_LIMITS: "true"` with `RATE_LIMIT_MULTIPLIER: "50"`, keeping rate limiters fully active while providing headroom for browser test runs.

4. **`DISABLE_REGISTRATION` Variable Dropped**:
   - Removed `DISABLE_REGISTRATION` from `server/storage.ts` and the ADR. The administrative toggle in the UI/API is the single source of truth.

5. **Route Limiters**:
   - Applied `publicReadLimiter` to `GET /api/system/public-settings` and `sensitiveActionLimiter` to `PUT /api/settings/system`.

6. **ADR Renumbered**:
   - Renamed `docs/adr/0006-system-settings-and-registration-policy.md` to `docs/adr/0007-system-settings-and-registration-policy.md` and updated its text.

7. **E2E Test Isolation**:
   - In `scripts/seed.ts`, `carol` is seeded with `emailVerified: false` to represent the unverified account fixture, so `POST /api/auth/login` rejects her with 403.
   - Followed Option 2: added a `finally` block in `tests/e2e/login-language.spec.ts` restoring Alice's language back to English (`"en"`).

8. **GitGuardian Annotations**:
   - Verified and ensured all test fixture credentials across touched files carry `// ggignore: throwaway test fixture credential`.

### Verification

- `npm run check` & `npm run check:sqlite`: Clean (0 errors)
- `npm run lint`: Clean (0 errors)
- `npm run db:check-drift`: Clean for both PostgreSQL and SQLite
- `npm test`: 38 / 38 files passed (498 passed, 9 skipped)
- `npm run test:e2e`: 33 / 33 passed

the-luap (commented):
Reviewed the rebase and `f099d1a`. All three blockers and the smaller items are fixed, and the two I could measure, I measured. One new thing to tighten, plus a collision with #35 that is nobody's fault but needs sorting before both land.

**Verified on the branch:**

- **The race is gone.** Same 8-way concurrent probe on an empty table that gave `1 ok, 7 rejected` before now gives `8 ok, 0 rejected`. Belt and braces too: the row is seeded in both migrations with `ON CONFLICT DO NOTHING`, the seeder inserts it, and the runtime fallback re-selects after a no-op insert. Any one of those would have done; having all three is fine.
- **`DISABLE_RATE_LIMITS` can no longer reach production.** Gated on `NODE_ENV === "test"`, which the e2e server does not set.
- **`DISABLE_REGISTRATION` is gone** — zero references in `server`, `shared`, `scripts`, `docs`. The admin toggle being the only control is the right call.
- Limiters on the new routes, ADR renumbered, and `login-language.spec.ts` now restores alice in a `finally` with the fetch wrapped, so a failure there cannot fail the spec.

## One thing to tighten

**`RATE_LIMIT_MULTIPLIER` is ungated and undocumented** — `server/utils/rate-limits.ts:7`

It is read straight from the environment at module load and multiplies all four limits, so `RATE_LIMIT_MULTIPLIER=1000` in production turns the login limiter into 15,000 attempts per fifteen minutes. That is a milder version of the thing we just closed: not an on/off switch, but the same shape, and nothing in `.env.example`, the README or the Synology guide mentions it exists. Give it the same `NODE_ENV === "test"` guard as `skipIfDisabled` and the knob does its job for the suite without being reachable on a real install.

Worth checking whether you still need it at all, or only for one limiter. `main` now counts **only failed** login attempts (#34, which landed after you branched), so the suite's sign-ins no longer consume that budget. What is likely still tight is `publicReadLimiter` at 120, because the login page calls the new `/api/system/public-settings` on every load and the suite loads a lot of pages. Scoping the headroom to that one would be narrower than scaling all four.

Minor, take it or leave it: `PUT /api/settings/system` puts the limiter **before** `authenticate`, where every other limited route in the codebase puts it after. Harmless, but it means unauthenticated probes spend the budget.

## Collides with #35

Not a defect in either PR, but they cannot both merge as they stand:

| | #33 | #35 |
| --- | --- | --- |
| Postgres migration | `0010_jazzy_korvac.sql` | `0010_colorful_the_executioner.sql` |
| SQLite migration | `0004_misty_guardsmen.sql` | `0004_safe_ezekiel.sql` |
| ADR | `0007-system-settings-…` | `0007-similarity-matching-…` |

Whichever merges second needs renumbering and a `meta/_journal.json` rebase. Since this one is green and small, I would land it first and renumber #35.

Everything else here looks ready to me.

Azumi0:
- **Rate limits**: Verified across the full E2E suite that `RATE_LIMIT_MULTIPLIER` is not needed. With #34 in place counting only failed logins and the whole browser suite generating well below the 120-request window on `/api/system/public-settings`, all 33 specs pass cleanly with standard limits. Removed `RATE_LIMIT_MULTIPLIER` from `server/utils/rate-limits.ts` and `playwright.config.ts`.
- **System settings route ordering**: Moved `sensitiveActionLimiter` after `authenticate, isAdmin` on `PUT /api/settings/system` in `server/routes/system-settings.ts` to align with all other admin routes and prevent unauthenticated probes from consuming the action quota.
- **Collision with #35**: Retaining the current migration numbers (`0010_jazzy_korvac.sql` / `0004_misty_guardsmen.sql`) and ADR `0007` on #33 so it can land first; #35 will be renumbered when addressed afterwards.

the-luap (commented):
Reviewed `01195c0`. Removing `RATE_LIMIT_MULTIPLIER` outright is a better answer than gating it, and the claim behind it holds up — I ran the full browser suite on this branch with the limiters fully active:

```
33 passed (41.9s)
```

`playwright.config.ts` now sets no rate-limit environment at all, so `skipIfDisabled` never fires during the run and every limiter is enforced at its production value. That is the strongest version of this: the suite passes against the same limits a real install uses, which means it will keep telling the truth if someone tightens them later.

Also verified:

- Zero references to `RATE_LIMIT_MULTIPLIER` anywhere in `server`, `client`, `playwright.config.ts` or `.env.example`.
- Limits back to 20 / 15 / 120 / 30, and `skipIfDisabled` still gated on `NODE_ENV === "test"`.
- `PUT /api/settings/system` is now `authenticate, isAdmin, sensitiveActionLimiter`, matching every other admin route. `GET /api/system/public-settings` keeps the limiter first, which is right — it has no auth to sit behind.

Agreed on keeping `0010` / `0004` / ADR `0007` here and renumbering #35 afterwards.

Nothing outstanding from my side. One cosmetic note for whenever you are next in that file: with nothing setting `DISABLE_RATE_LIMITS` any more, `skipIfDisabled` is dead code. Worth either dropping it or leaving a line saying it is a deliberate test-only escape hatch for when the suite outgrows the limits — no need to hold the PR for it.

PR #35, opened 2026-09-11
feat: similarity matching, catalog auto-registration, and admin-managed generic terms

Azumi0:
## Description

This PR addresses real issues I've faced when trying to use the app - spool onboarding edge cases and catalog hygiene when importing filaments from barcodes, NFC tags, and community databases (Open Filament Database & SpoolmanDB), introducing:

1. **Auto-Registration of Unknown Entities on Save (ADR-0007)**
   - Automatically registers unknown manufacturers into the `manufacturers` table (`ensureDeclaredManufacturerResolves`) case-insensitively.
   - Automatically registers novel materials into the user's Personal Catalog (`materials`), persisting scanned density values.
   - Handles concurrent insert race conditions by querying the canonical casing if an insert was skipped due to conflict.

2. **Multi-Tier Fuzzy Similarity Matching (ADR-0007)**
   - Implements `findSimilarItems` in `shared/similarity.ts` using Unicode-aware alphanumeric normalization, significant token overlap, prefix containment, and Levenshtein edit distance `<= 2`.
   - Prompts users sequentially in `filament-modal.tsx` when scanned metadata closely matches existing catalog entities, preventing accidental near-duplicate entries while preserving user choice.
   - Employs symbol-aware word boundaries (`\b` matching for trailing symbols like `PLA+`) and safe function replacers to prevent substring corruption in generated spool names.

3. **Admin-Managed Generic Terms (ADR-0008)**
   - Adds a `generic_terms` table with dual-dialect migrations (PostgreSQL and SQLite) to filter out generic industry noise words (e.g. "Lab", "Filament", "3D") from token overlap matching.
   - Admin settings tab (`<GenericTermsList />`) with CSV import/export, role-based access control, and complete localization across English, German, and Polish.
   - Starter seeding of 12 default terms (`lab`, `labs`, `filament`, `filaments`, `3d`, `polymers`, `material`, `materials`, `printing`, `print`, `studio`, `maker`) with DB sequence inspection to prevent re-seeding if an admin deliberately clears the table.

4. **UI Deduplication & Color Alignment**
   - Filters out predefined material options in dropdowns when they already exist in the user's database.
   - Relabels the color selector to "Color Template (optional)" and makes the Hex Code input mandatory with template autofill.

5. **Responsive Mobile Chart Presentation**
   - Replaces floating SVG text slice labels on viewports `< 768px` in `material-color-chart.tsx` with a responsive flex-wrap badge list.

6. **Architecture Decision Records & Documentation**
   - Documents decisions in `docs/adr/0007-similarity-matching-and-catalog-auto-registration.md` and `docs/adr/0008-admin-managed-generic-terms.md`.
   - Updates `CONTEXT.md` domain glossary with "Generic Term".

## Type of change

- [x] Documentation update
- [x] Code refactoring

## How Has This Been Tested?

- [x] **Unit & Component Tests**: 
  - `tests/utils/similarity.test.ts` (17 tests covering exact, token, prefix, Levenshtein, and stop word filtering)
  - `tests/components/filament-modal.test.tsx` (13 tests verifying modal state, symbol replacement, and template sync)
  - `tests/components/material-color-chart.test.tsx` (responsive chart rendering tests)
  - `tests/components/settings-generic-terms.test.tsx` (admin settings tab and full i18n parity tests)
- [x] **API & Integration Tests**:
  - `tests/routes/generic-terms.test.ts` (CRUD, authentication, authorization, CSV export/import, sequence inspection)
  - `tests/routes/filaments.test.ts` (concurrent insert resolution test with mixed casing)
- [x] **E2E & Type Checks**:
  - `tests/e2e/mobile.spec.ts` & `tests/e2e/community-catalog-add-modal.spec.ts`
  - `npm run check` & `npm run check:sqlite` (0 type errors)
  - Full test suite: 38 test files, 539 tests passed, 0 failures.

## Checklist:


the-luap (commented):
Lots of genuinely useful work here — the similarity prompt is a real answer to scan-created near-duplicates, and generic terms are correctly admin-gated through the existing CRUD factory. Two things block it, one of which I reproduced against a running instance, and there is a collision with #33 to sort out.

## Blocking

**1. Saving a spool now writes to the admin-only global manufacturer catalog** — `server/storage.ts:231` (`ensureDeclaredManufacturerResolves`, called from `findOrCreateFilamentType`)

`manufacturers` is not user-scoped, so `registerCrudSettingsRoutes` guards it with `[authenticate, isAdmin]` — creating one has always been an admin action. `findOrCreateFilamentType` runs on **every** filament create, for any authenticated user, and inserts unknown names straight into that shared table. Verified on the branch against a seeded instance, as `alice` (role `user`):

```
POST /api/manufacturers            -> 403        (still refused, as designed)
POST /api/filaments  manufacturer="AliceCorp-ViaSpool" -> 201
GET  /api/manufacturers  (as admin) -> "AliceCorp-ViaSpool"   # now global, 3 rows -> 4
```

So the feature routes around the gate: a non-admin cannot add a manufacturer, but can add one by naming it on a spool, and it appears in every user's dropdown with no review. The CSV and JSON import paths go through the same `storage.createFilament`, so one request can add up to 2000 of them.

The materials half of this change is right and shows the contrast: `ensureDeclaredMaterialResolves(userId, ...)` inserts with `userId`, i.e. into that user's **Personal** Catalog, which is exactly the per-user scope ADR 0003 established. Manufacturers have no owner column, so the same trick is not available — which is the point. Options, roughly in order of how much I like them:

- Resolve against the catalog but **do not insert**: keep the spool's free-text manufacturer as it is today, and let the similarity prompt do the deduplicating. The spool already stores the name; nothing is lost.
- Give `manufacturers` an owner column the way `materials` has one, and auto-register into the user's own scope. Bigger change, but consistent.
- Keep the global insert but make it admin-only, and for non-admins fall back to the free-text name.

Whichever you pick, the ADR should say plainly that this widens who can write a shared table, because right now it reads as a convenience feature.

**2. `lint` is red** — two `no-extra-boolean-cast` errors, `scripts/seed.ts:108` and `tests/routes/generic-terms.test.ts:269`. `eslint --fix` handles both.

## Smaller

**The race comment in `ensureDeclaredManufacturerResolves` describes the wrong race.** It says a concurrent insert with different casing would be "silently skipped" by `onConflictDoNothing`, but `manufacturers_name_key` is a plain case-sensitive unique index on both engines, so inserting `prusa` while `Prusa` exists does **not** conflict — it succeeds, and you end up with two rows that the `lower()` lookup then picks between arbitrarily. Sequentially the lookup prevents this; only true concurrency reaches it. Either fold case in the index (a migration) or say in the comment that case-variant duplicates are possible under concurrency.

## Collides with #33

Both PRs claim the same numbers:

| | #33 | #35 |
| --- | --- | --- |
| Postgres migration | `0010_jazzy_korvac.sql` | `0010_colorful_the_executioner.sql` |
| SQLite migration | `0004_misty_guardsmen.sql` | `0004_safe_ezekiel.sql` |
| ADR | `0007-system-settings-…` | `0007-similarity-matching-…` |

#33 is green and small, so I would land that first and renumber here — this PR's ADRs become 0008 and 0009, and the migrations shift by one with a `meta/_journal.json` rebase.

## Checked and fine

Similarity runs only in `filament-modal.tsx`, so the Levenshtein cost is client-side over a small catalog and not a server concern. Generic terms go through the same factory as the other non-scoped entities and inherit the admin guard, CSV import included. The seed guard's sequence check is more machinery than I would reach for, but it does what it says and is dialect-aware.

the-luap:
A second pass over the diff turned up three things my review above missed, one of them worse than anything I raised the first time. All three are verified on the branch.

## Blocking

**A. The re-seed guard is dead code on SQLite, so deleted generic terms come back on every restart** — `scripts/seed.ts:95`

The guard calls `(db as any).execute(...)`. The SQLite driver is `LibSQLDatabase`, which exposes `run / all / get / values / transaction / batch` and **no `execute`** — the `as any` is what hid that from `tsc`:

```
execute: undefined | get: function | all: function | run: function
```

So the call throws `TypeError`, the bare `catch {}` swallows it, and control falls through to the insert. End to end on the branch, SQLite:

```
after migrate:                 0
after first seed:             12
after admin deletes them:      0
after restart / second seed:  12     <- and nothing logged
```

That is precisely the case ADR-0008 says the guard exists to prevent, on the dialect the homelab and Synology installs actually run. An admin who removes `filament` or `3d` because it hurts their matching gets them back at the next container restart, silently, forever. The Postgres branch is fine — `node-postgres` does have `execute`. Use `db.get(...)` or `db.all(...)` for the SQLite branch, and consider logging in the `catch` rather than swallowing, which is what let this hide.

**B. `density` reaches a numeric column with no format check** — `shared/schema.ts:395`

`density: z.union([z.string(), z.number()]).nullable().optional()` accepts anything, and it flows into `ensureDeclaredMaterialResolves` → `db.insert(materials).values({ density })` with no `try`. On the branch:

```
POST /api/filaments  material="NovelMat1"  density="abc"   -> 201
materials row:  [{"name":"NovelMat1","density":"abc"}]
```

On SQLite `t.numeric` is TEXT, so that is stored and every later weight/length calculation on it yields `NaN`. On Postgres the same insert is `invalid input syntax for type numeric` and an unhandled 500. `updateMaterialSchema` already guards this exact hole with `z.string().regex(/^\d+(\.\d+)?$/)` and its comment calls out the 500 by name; the new field needs the same.

## Should fix

**C. The mandatory `#RRGGBB` colour code blocks editing spools that already exist** — `client/src/components/filament-modal.tsx:176`

`colorCode` goes from optional to `.min(1).regex(/^#[0-9A-Fa-f]{6}$/)`, while the server still accepts `z.string().max(20)` and the community catalog builds it as `hex.startsWith("#") ? hex : "#" + hex` with no length check on the upstream value. A spool imported from an OFD or SpoolmanDB entry with a 3-digit or 8-digit hex, or a CSV row with `#FFF`, now fails validation the moment its owner opens the modal to change anything at all — remaining percentage included — until they retype the colour by hand. Either relax the rule to what the server accepts, or normalise on read.

## Smaller, worth a look

- `server/routes/settings.ts:226` — `isInUse: () => false` with no `userScoped` sends every generic-term delete down the global branch, which calls `getFilamentsOfAllUsers()` and loads the whole cross-user filament table to evaluate a predicate that is constantly false. "Delete all" does it once per term.
- `client/src/components/settings/settings-crud-list.tsx:45` — widening `entityType` to `CatalogRequestEntityType | string` collapses to `string`, so a typo in any other list's entity type now compiles. The Generic Terms tab never sends one, so leaving it `undefined` would keep the union intact for everyone else.
- `scripts/seed.ts:354` — `seedGenericTerms()` is only wired into the non-demo branch, so a `--demo` install starts with an empty stop-word list and the similarity prompt false-positives on exactly the words the feature exists to suppress.

Together with the manufacturer finding above, that is two blockers and one behaviour regression. Happy to re-check once they are in.

the-luap:
#33 is merged, so this branch now conflicts. What it needs on top of the findings already raised:

**Renumbering**, as agreed on #33, which kept `0010` / `0004` / ADR `0007`:

| | now | should become |
| --- | --- | --- |
| Postgres migration | `0010_colorful_the_executioner.sql` | `0011_…` |
| SQLite migration | `0004_safe_ezekiel.sql` | `0005_…` |
| ADRs | `0007-similarity-matching-…`, `0008-admin-managed-generic-terms.md` | `0008`, `0009` |

Both `meta/_journal.json` files need the rebase too, and `npm run db:check-drift` will confirm it.

Still open from the two reviews above: the global manufacturer write from a non-admin spool save, the dead SQLite re-seed guard, unvalidated `density` reaching a numeric column, the mandatory six-digit colour code blocking edits of existing spools, and the red `lint` check (two `eslint --fix`-able errors).

the-luap (commented):
Reviewed `1faba42`. Every finding from both earlier rounds is fixed, and I verified the four substantive ones against a running instance rather than the diff. Taking the deepest option on manufacturers was the right call. One new blocker came in with that fix, and `e2e` is still red for three reasons that are all spec-level.

## Verified fixed

- **Manufacturer scoping.** Mirrors materials exactly: nullable `userId`, partial unique indexes for the global and per-user halves. Live, as `alice` (role `user`): `POST /api/filaments manufacturer="AliceOnlyBrand"` → 201, the name appears in **alice's** list and **not** in the admin's. That closes the bypass.
- **Seed guard.** `db.all` instead of the non-existent `db.execute`, and the `catch` logs now. End to end on SQLite: seed → 12, admin deletes → 0, restart → **0**. Previously 12.
- **Density.** `POST` with `density: "abc"` → `400 Density must be a positive number`, where it was a 201 storing `"abc"`.
- **Colour code.** Relaxed to `z.string().max(20).optional().nullable()`, matching the server, with 3/4/8-digit hex normalised on read.
- Also confirmed: the `CatalogRequestEntityType` union is restored, `isInUse` is optional so generic-term deletes no longer scan every user's spools, the demo seed includes generic terms, `lint` is green, and the renumbering plus `db:check-drift` are clean.

## New blocker: the SQLite half of that migration has no guard

`migrations/pg/0012_slow_plazm.sql` opens with a `DO $$` block that aborts with a clear message when case-insensitive duplicates exist. `migrations/sqlite/0006_nappy_marvex.sql` has no equivalent — it drops the old index and creates the new unique ones directly.

That matters because case-variant manufacturers are a state this project **deliberately permits**. #19 made the duplicate check exact on purpose, and `tests/routes/settings.test.ts` on `main` still asserts `POST /api/manufacturers {name:"prusa"}` after `"Prusa"` returns **201**. So installs in the wild have them, and this PR is the first thing to disallow it.

Simulated the upgrade: built `main`, migrated a fresh SQLite database, inserted `Prusa` and `prusa`, then ran this branch's migrator:

```
SQLITE_CONSTRAINT_UNIQUE  (rawCode 2067)
rows: query failed - no such column: user_id
applied migrations: 5
```

A raw constraint error, no indication of which names collide, and the database left half-migrated. The entrypoint runs migrations before the app starts, so that is a crash loop on exactly the SQLite homelab and Synology installs this series has been about. `verify-upgrade` misses it because its fixture has no case-variant names.

Two ways to fix it, and the second is available here in a way it usually is not:

1. Mirror the Postgres guard. SQLite has no `RAISE` outside a trigger, so the natural home is a pre-flight check in `scripts/migrate.sqlite.ts` that refuses with the same message before `migrate()` runs.
2. **Deduplicate in the migration instead of failing.** Nothing foreign-keys to `manufacturers.id` — a spool's manufacturer is a text column on `filament_types` — so these rows are only a picklist, and collapsing `prusa` into `Prusa` orphans nothing. That turns a failed upgrade into a silent, correct one. Worth considering for Postgres too, since the exception there is also a dead end for the operator.

Either way, please add a case-variant pair to the `verify-upgrade` fixture so CI covers it.

## Still red: `e2e`, three failures

All spec-level, none of them the feature's fault, but they need fixing before this lands:

1. **`filament-modal-barcode-field.spec.ts:33`** — clicks `getByLabel(/color\*/i)`, and this PR relabels that field to "Color Template". The spec passes on `main`, so this is a coverage regression introduced here.
2. **`mobile.spec.ts:86`** — expects `/list management/i`, which #33 renamed to "General Settings". `tests/e2e/helpers.ts` already matches both spellings; this new spec matches only the old one.
3. **`mobile.spec.ts:51`** — strict-mode violation: `getByTestId('mobile-chart-labels').getByText('PLA')` resolves to two nodes, because the category and its percentage share one `<span>`. `{ exact: true }` or a narrower locator fixes it.

## Minor

The submit handler backfills an empty colour name with the colour code, falling back to the literal `"Color"` (`filament-modal.tsx:502`). It prevents a 400, but it papers over the client schema marking `colorName` optional while the server still requires it, and it can leave a spool whose colour is named `#000000`. Making the two schemas agree would be cleaner than normalising at the boundary.

Azumi0:
All points from the latest review have been addressed and verified in commits `4f59415` and `4014b35`:

### 1. Migration Deduplication on Upgrade (SQLite & PostgreSQL)
- **Automatic Deduplication in Migrations**: Implemented Option 2 in both PostgreSQL (`migrations/pg/0012_slow_plazm.sql`) and SQLite (`migrations/sqlite/0006_nappy_marvex.sql`). The migrations run a pre-index deduplication query (`DELETE FROM manufacturers WHERE id NOT IN (SELECT min(id) FROM manufacturers GROUP BY coalesce(user_id, -1), lower(name))`) before creating the case-insensitive unique indexes. This collapses legacy case-variant duplicates (e.g. `prusa` into `Prusa`) cleanly without operator intervention or crash loops.
- **Upgrade Verification**: Added a case-variant collision pair (`('Prusa', 4), ('prusa', 5)`) to `scripts/verify-upgrade.ts` and asserted that `prusa` is collapsed into `Prusa`.
- **SQLite Upgrade Regression Test**: Added `tests/migrations-sqlite-dedup.test.ts` to simulate upgrading a SQLite database created under the pre-0006 schema with case-variant collisions.

### 2. Resolved E2E Regressions & Schema Parity
- **Color Label & Schema Alignment**: Restored `<FormLabel>{t('filaments.color')}*</FormLabel>` (`Color*`) and required client schema `colorName: z.string().min(1, ...)`. Removed the boundary fallback hack (`colorName || colorCode || 'Color'`), ensuring the client and server schemas strictly agree.
- **Mobile E2E Selectors**:
  - Updated `tests/e2e/mobile.spec.ts` to use `/general settings|list management/i` to handle the post-#33 settings menu renaming.
  - Added `{ exact: true }` to the mobile chart label text locator to prevent Playwright strict-mode ambiguity.
- All 39 Playwright E2E tests pass cleanly.

### 3. Density Validation & Storage Safety
- **Strict Positive Validation**: Added `.refine((v) => Number(v) > 0)` to `filamentWriteSchema` and the client form schema, rejecting `"0"`, `"0.0"`, and negative numbers with 400.
- **Storage Protection**: Added `Number(rawDensity) > 0` validation to `ensureDeclaredMaterialResolves`. Destructured `density` out in `DatabaseStorage.updateFilament` so virtual material fields do not leak into the `filaments` table update, and guarded against empty updates.
- **UI Division Guard**: Added defensive numeric checks in `calculateRemainingLength` to prevent division by zero or NaN.

### 4. Similarity Matching Refinements
- **Prefix Match Threshold**: Raised prefix containment minimum clean length to 3 (`cleanScanned.length >= 3 && itemClean.length >= 3`) so 2-letter codes like `PE` do not match `PETG`.
- **Glossary Alignment**: Renamed `stopWords` parameter to `genericTerms` in `shared/similarity.ts` and updated ADR-0008.

### 5. Verification Results
- `npm run lint`: 0 errors, 0 warnings
- `npm run type-check`: 0 errors
- `npm run check:sqlite`: 0 errors
- `npm run db:check-drift`: Passed (0 drift across PostgreSQL & SQLite)
- `npm run db:verify-upgrade`: Passed (all assertions pass, case-variant duplicates deduplicated)
- `npm test`: Passed (43 test files, 565 passed, 0 failed, 9 skipped)
- `npm run test:e2e`: Passed (39/39 passed)

the-luap (commented):
Reviewed `4f59415` and `4014b35`. Everything is addressed, all ten checks are green, and I re-ran the two things I had reproduced as failures. Ready to merge from my side.

**The upgrade blocker is gone.** Same simulation as before — built `main`, migrated a fresh SQLite database, seeded case-variant names, then ran this branch's migrator:

```
before:  1:Prusa, 2:prusa, 3:Bambu Lab, 4:BAMBU LAB, 5:Sunlu
after:   1:Prusa(u=null), 3:Bambu Lab(u=null), 5:Sunlu(u=null)
migrations applied: 7      exit 0
```

Statement order is right in both dialects: the column is added before the `DELETE` references it, and the `DELETE` runs before the indexes are created. Taking option 2 and deduplicating rather than aborting was the better call — an operator who upgrades never sees this at all, and nothing references these rows by id so no spool is affected. Adding the collision pair to `verify-upgrade` means CI now covers it.

**Density**, live: `2.5` → 201, `0` → 400, `-1` → 400, `"abc"` → 400.

**Colour**, back in agreement: the label is `Color*` again and `colorName` is required on both sides, with the boundary fallback removed — so no more spools coloured `#000000`, and `filament-modal-barcode-field.spec.ts` passes again.

**The `updateFilament` leak you found on your own** is a good catch: `PATCH` with `density` now answers 200 instead of pushing a virtual material field at the `filaments` table, and an empty body is guarded.

`npx playwright test` — **39 passed** locally, including both mobile specs.

One cosmetic note for later, not worth holding this: the dedup keeps `min(id)`, so the earliest-inserted casing survives. On an install where the lowercase row happens to be older, the picklist will read `prusa` while existing spools still carry `Prusa` in their own text column. Nothing breaks, and an admin can rename it.

PR #36, opened 2026-09-12
feat: custom manufacturer & material persistence controls (ADR-0010), modal reset lifecycle, and QR parsing fixes

Azumi0:
## Description

This PR introduces user-controlled catalog persistence for custom manufacturers and materials ([ADR-0010](docs/adr/0010-custom-manufacturer-material-persistence-controls.md)), fixes filament modal state reset and lifecycle leak issues, improves barcode and QR scanning accuracy, and resolves SQLite starter data re-seeding bugs on container restarts.

### 1. User-Controlled Persistence & Custom Inputs (ADR-0010)
- **Manufacturer Selection**:
  - Added dedicated `"Other"` dropdown option. Selecting "Other" (or scanning/importing an uncataloged manufacturer) displays a custom text input field.
  - Added a `"Save manufacturer to my collection"` persistence checkbox beneath the custom input (defaults to checked `true`). Unchecking allows users to add spools from temporary or one-off manufacturers without polluting their Personal Catalog in Settings.
- **Material Selection**:
  - Added dedicated `"Custom"` dropdown option. Selecting "Custom" (or scanning/importing an uncataloged material) reveals a custom material name input field.
  - Added a `"Save material to my collection"` persistence checkbox beneath the custom input (defaults to checked `true`).
- **Decoupled Backend Storage**:
  - Updated `shared/schema.ts` with `saveManufacturer?: boolean` and `saveMaterial?: boolean` in `filamentWriteSchema`, `filamentPatchSchema`, and `InsertFilament`.
  - Updated `server/storage.ts` (`findOrCreateFilamentType`): when `saveManufacturer: false` or `saveMaterial: false`, catalog row insertion into `manufacturers` / `materials` tables is bypassed while the string values remain safely persisted on `filament_types`.
  - Full backward compatibility: omitting the flags (e.g. in CSV / Spoolman imports) continues to default to `true`.
- **i18n**: Added localized strings in English, Polish, and German (`filaments.otherManufacturer`, `filaments.customManufacturerName`, `filaments.saveManufacturerToCollection`, `filaments.customMaterialName`, `filaments.saveMaterialToCollection`).

### 2. Filament Modal State Reset & Lifecycle Management
- Completely reset form values and local state (`customManufacturerName`, `customMaterialName`, `customColorName`, persistence checkboxes, candidate prompt, history view) when the modal is closed, cancelled, or saved.
- Synchronized `field.onChange` and form controllers across all dropdown search inputs (`onKeyDown` Enter handler now triggers matching and custom toggling for Color, Manufacturer, and Material).
- Cleared parent copy state (`selectedFilament`, `copyFromFilament`) in `client/src/pages/home.tsx` on modal open/close/save to prevent stale data leaks across successive modal interactions.
- Maintained exit transitions by closing through Radix dialog events instead of abruptly unmounting components.

### 3. Material Deduplication & Canonicalization
- `uniqueMaterialOptions` implements symmetric, order-independent deduplication prioritizing shorter canonical names (e.g. `PA` over `PA (Nylon)`) without dropping distinct types (like `PLA Silk`) or dropping parenthesized entries.
- Canonicalized `colorName` in form `defaultValues`, `form.reset`, and async `useEffect` when catalog colors load to ensure exact catalog match.
- Reserved `"Custom"` keyword excluded from dynamic color/material lists to prevent duplicate Select item keys.

### 4. Barcode & QR Scanner Enhancements
- Stopped prepending manufacturer to filament names during barcode and QR scanning (e.g. Bambu Lab spools no longer result in "Bambu Lab Bambu Lab PLA").
- Restored token-based color extraction, modifier retention (e.g., 'Matte'), and CF/HF material detection without false serial matches; added PETG-HF support.
- Fixed ESLint no-useless-assignment warnings in the scanner component.
- Auto-matches recognized scanned colors to catalog entries and falls back cleanly to "Custom" when unrecognized.

### 5. Database Seeding & SQLite Restart Guard
- Added checks in `scripts/seed.ts` (Check 1: `system_settings.id = 1`, and Check 3: sequence check in `sqlite_sequence` / `pg_sequences`) to prevent starter data re-seeding on reboot when tables were intentionally cleared.

## Type of change

- [x] Documentation update
- [x] Code refactoring
- [ ] Performance improvement

## How Has This Been Tested?

- **Unit Tests**:
  - `tests/components/filament-modal.test.tsx`: 40 tests covering custom manufacturer/material inputs, persistence checkboxes, color matching, search input Enter handling, and parenthetical strip deduplication.
  - `tests/components/qr-scanner.test.ts`: QR code parsing, color extraction, and material detection across Bambu Lab variants.
- **Integration Tests**:
  - `tests/routes/filaments.test.ts`: verified `saveManufacturer: false` and `saveMaterial: false` API behavior with decoupled storage.
  - `tests/seed-sqlite-restart.test.ts`: verified starter re-seed prevention after database clearing.
- **E2E Tests**:
  - `tests/e2e/filament-modal-reset.spec.ts`: verified form reset on close, cancel, and save without state leak.
- **Full Suite Verification**:
  - All 45 test files passed (599 passing tests, 9 skipped).
  - `npm run lint` passed with 0 warnings.
  - `npm run check` (TypeScript typecheck) passed with 0 errors.

## Checklist:

- [x] I have made corresponding changes to the documentation (added ADR-0010)


the-luap (commented):
The persistence controls are a good answer to catalog pollution, and the flags are plumbed correctly — I checked the thing that bit the last PR and `saveManufacturer`/`saveMaterial` are destructured out in both `createFilament` and `updateFilament`, so they never reach the `filaments` insert. One finding blocks it, verified against a real Postgres container.

## Blocking: `INIT_SAMPLE_DATA=true` silently does nothing on Postgres

Check 1 treats "a `system_settings` row with id 1 exists" as evidence that starter data was already seeded. But that row is written by a **migration** (#33's `0010_jazzy_korvac.sql`), not by seeding, so on Postgres it always exists by the time the seeder runs — the entrypoint migrates first, then seeds. Live, on a fresh `postgres:16-alpine`:

```
after migrate, before seeding:   system_settings=1
seed --starter:                  "System settings already initialized, skipping starter initialization."
result:  manufacturers=0  materials=0  colors=0  storage_locations=0
```

A fresh Postgres install that opts into starter data gets none of it, with a log line that reads like success.

**It only works on SQLite by accident.** Neither migration has a `--> statement-breakpoint` between the `CREATE TABLE` and the `INSERT`, so libsql executes just the first statement and the insert is silently dropped — `system_settings` is empty after migrating, Check 1 does not fire, and seeding proceeds. Verified:

```
sqlite,  after migrate:  system_settings rows: 0     -> starter seed runs, 4 manufacturers
postgres, after migrate: system_settings rows: 1     -> starter seed skipped entirely
```

So the guard's behaviour currently depends on a migration bug, in opposite directions per dialect.

Two things to fix, I think:

1. **Drop Check 1.** Checks 2 and 3 already answer the real question — rows present, or a sequence that has advanced. `system_settings` says nothing about whether the starter catalog was ever seeded, because nothing in the seeder creates it.
2. **Add the missing `--> statement-breakpoint`** to both `system_settings` migrations so the seeding insert actually runs on SQLite too. Harmless today, since `getSystemSettings()` creates the row lazily with `onConflictDoNothing`, but it is a latent trap and it is what makes the two dialects disagree here. That one is already on `main`, so it can be a separate commit if you prefer.

**CI cannot catch this as written.** The `seed-postgres` and `seed-sqlite` jobs run `db:migrate` then `db:init` and check the exit status only; they never assert that any rows landed. A one-line count assertion after the starter fixture would have caught it, and would be worth adding alongside the fix.

## Verified working

- `saveManufacturer: false` + `saveMaterial: false` → spool created, the strings stored on the spool, and **nothing** added to the user's catalogs. Omitting both still saves, so CSV and Spoolman imports keep their old behaviour.
- The re-seed guard for the case it was written for: seed, clear the lists as an admin, restart → they stay cleared.
- ADR numbering is clean at 0010.

I have not gone through the modal lifecycle and QR parsing changes in detail yet — they are the bulk of the diff and the unit tests cover them; happy to look closer if you want a second opinion on any part.

the-luap:
`e2e` has finished and is red: **3 failed, 40 passed**. All three are your own naming change meeting specs that still assert the old behaviour, so they are spec updates rather than defects — but they need doing before this lands.

| spec | expected | got |
| --- | --- | --- |
| `filament-modal-scanner.spec.ts:120` | `/Galaxy Silver.*Prusament/i` | `PLA Galaxy Silver` |
| `community-catalog-add-modal.spec.ts:46` | (name containing the brand) | `PolyTerra PLA Sapphire Blue` |
| `community-catalog-add-modal.spec.ts:179` | (name containing the brand) | `Galaxy Black` |

That is exactly what "stopped prepending manufacturer to filament names" is supposed to produce, and the new names read better. Both specs come from #32/#35 and pass on `main`, so they simply encode the old rule and should be updated to the new one.

Worth deciding explicitly while you are there: `PLA Galaxy Silver` keeps the material in the name but drops the brand, whereas `Galaxy Black` has neither. If the intended rule is "colour and material, never the manufacturer", the second one looks like it lost the material too — worth a glance at whether that is the rule or an artifact of which fields the community entry happened to carry.

Everything else is green, including both dialect suites, drift and lint. The seeding finding in my review above still stands.

the-luap:
A second pass found four more, all in the scanner and modal code I said I had not gone through. Every one reproduced on the branch by executing the real functions. Two of them are user-visible wrong data.

## Blocking

**A. Scanning a Bambu barcode reports the wrong colour** — `qr-scanner.tsx:108`

`matchColorCode` uses unanchored `includes()` on two-letter codes, so `BL` hits inside `BLACK` and `OR` inside `IVORY`. Run against the branch:

```
BBL-PLA-BLACK  -> colorName "Blue"   colorCode #0A2989
BBL-PLA-IVORY  -> colorName "Orange" colorCode #FA6607
```

Scanning a black spool fills the form in as blue. The `parts[2]` fallback that would have handled the spelled-out name is unreachable once a substring hits. Anchor the codes, or try the full-word match first.

**B. PCTG parses as PC** — `qr-scanner.tsx:227`

`upper.startsWith('PC')` is tested before `startsWith('PCTG')`, so the PCTG branch can never run:

```
[BBL]PCTG Black 1KG  -> material "pc", printTemp "260-290", name "PC Black"
```

It should be `pctg` at 250-270. The sibling `processBambuLabBarcode` orders these correctly, so it is just this chain. `PETG-HF` works, for what it is worth.

**C. Opening a spool whose material has a parenthetical silently rewrites it on save** — `filament-modal.tsx:485`

`uniqueMaterialOptions` drops any entry whose stripped form was already seen, and `findMatchingMaterial` falls back to the stripped comparison. With a catalog of `PLA`, `PLA (Silk)`, `PA`, `PA (Nylon)`:

```
dropdown shows:                              ["PA", "PLA"]
findMatchingMaterial("PLA (Silk)", ["PLA"])  -> PLA
```

So `PLA (Silk)` is invisible in the picker, and `form.reset` resolves an existing `PLA (Silk)` spool to `PLA`. Change the remaining percentage, save, and the spool's material has quietly become `PLA`. `findMatchingManufacturer` behaves the same way with `Prusa (Research)`. The test that motivates this dedupe is about annotation labels like `PA (Nylon)`, but it is applied to arbitrary user catalog entries where the stripped form is a different product. Since #35 auto-registers whatever a user types, those entries are easy to end up with.

**D. A scan with no recognisable colour clears the colour already chosen** — `filament-modal.tsx:1017`

The guard widened from `if (data.colorName)` to `if (data.colorName || data.colorCode)`, and both Bambu parsers always return a colour code even when nothing was detected:

```
BBL-PLA-901234 -> colorName undefined, colorCode "#000000"
```

`applyColorData(undefined, '#000000')` then takes the no-match branch, flips the field to Custom and sets `colorName` to `""` — wiping whatever the user had picked and leaving a required field empty.

## Also on the seeding, confirming my earlier point

The Postgres branch of Check 3 cannot work either — `pg_sequences` has no `is_called` column:

```
ERROR:  column "is_called" does not exist
columns: last_value, increment_by, cycle, cache_size, data_type, start_value,
         min_value, max_value, sequencename, sequenceowner, schemaname
```

It is inside `try/catch`, so it degrades to a `console.warn` and falls through. Line 170 of the same file gets this right by selecting from the sequence relation directly. Dropping `is_called` and testing `last_value > 1` would work, since `pg_sequences.last_value` is NULL on an untouched sequence.

So on Postgres both the new guards are inoperative: Check 1 for the wrong reason, Check 3 because the query throws.

---

Together with the earlier review that is one starter-seeding regression, two guards that do not work, two scanner bugs producing wrong values, one silent material rewrite, and three e2e specs to update. Happy to re-check whenever it is ready.

Azumi0:
All review findings and edge cases have been resolved in the latest commits (`8e5358a..8c8c5c3`):

### 1. Bambu Barcode Color Matching (`qr-scanner.tsx`)
- Separated color matching into full-word matches (`COLOR_NAME_MAP`) and anchored two-letter tokens (`TWO_LETTER_MAP`).
- Full-word colors are checked first, and two-letter codes are anchored to token delimiters (`(^|[-_/ ])CODE([- _/]|$)`). `BBL-PLA-BLACK` now matches `Black` (`#000000`) and `BBL-PLA-IVORY` matches `Ivory` (`#F5F5F0`) without false hits on `BL` (Blue) or `OR` (Orange).
- When no color is recognized, scanners return `colorCode: undefined` instead of defaulting to `#000000`.

### 2. PCTG Parsing Order (`qr-scanner.tsx`)
- Reordered `PCTG` before `PC` in both `upper.startsWith(...)` checks and the `materialRegex` alternation (`(PLA|ABS|PETG|TPU|PA|ASA|PCTG|PC|...)`).
- Added `'pctg': 'PCTG'` to `materialMap`. `[BBL]PCTG Black 1KG` now parses as `pctg`, print temp `250-270`, and name `PCTG Black`.

### 3. Parenthetical Material & Manufacturer Retention (`filament-modal.tsx`)
- Reverted parenthetical stripping from dropping distinct catalog items in `uniqueMaterialOptions` and from silently rewriting items in `findMatchingMaterial` and `findMatchingManufacturer`. Distinct items such as `PLA (Silk)` and `Prusa (Research)` are preserved in dropdowns and never rewritten to base names on edit/save.
- Fixed `uniqueMaterialOptions` fallback suppression so custom database variants like `PLA (Silk)` do not suppress the predefined standard `PLA` fallback when standard `PLA` is not yet in the database.

### 4. Color Selection Preservation (`filament-modal.tsx`)
- Scans or templates with no recognizable color now return `undefined`, and `applyColorData` leaves the user's currently selected color untouched rather than switching to Custom `#000000`.

### 5. Seeding & Migrations (`scripts/seed.ts`, migrations)
- Added `--> statement-breakpoint` between `CREATE TABLE` and `INSERT INTO` statements in `0010_jazzy_korvac.sql` (Postgres) and `0004_misty_guardsmen.sql` (SQLite) so multi-statement Drizzle migrations run reliably on SQLite.
- Removed Check 1 (`system_settings`) in `scripts/seed.ts` so `INIT_SAMPLE_DATA=true` runs on fresh Postgres databases.
- Fixed Check 3 on Postgres: removed the nonexistent `is_called` column and test `last_value IS NOT NULL AND Number(last_value) > 1` (untouched sequences have `last_value IS NULL` or 1; seeded starter data generates 10+ rows).
- Added starter row count assertions (`COUNT(*) > 0`) to `.github/workflows/test.yml` for both `seed-sqlite` and `seed-postgres`.

### 6. Community Catalog Naming & E2E Specs
- Implemented Option 2 naming (`${material} ${baseName}`) without appending manufacturer name, centralized in `client/src/lib/community-catalog.ts`.
- Uses regex word boundary checks (`\b`) to prevent false substring matches on 2-letter materials (`PA`, `PP`) and short colors (`Red`, `Tan`) inside product names (e.g. `Sparkle Blue`, `Copper`, `Shredder`).
- Updated the 3 E2E test assertions in `community-catalog-add-modal.spec.ts` and `filament-modal-scanner.spec.ts`.
- Prefills the scanned code into `Barcode / GTIN` on the Add Filament modal even when a scanned barcode has no catalog match or lookup fails.

### Verification
- **Unit tests**: All 46 test files (617 tests) passing (`vitest run`).
- **E2E tests**: All 8 specs passing (`playwright test tests/e2e/community-catalog-add-modal.spec.ts tests/e2e/filament-modal-scanner.spec.ts`).
- **Linters & Drift**: `npm run db:check-drift`, `npm run check`, `npm run check:sqlite`, and `npm run lint` all pass with 0 errors/warnings.

the-luap (commented):
Reviewed `a8bb8ec..8c8c5c3`. Everything from both rounds is fixed, and I re-ran each thing I had reproduced as broken. Ready to merge from my side.

**Scanner**, executing the real parsers on the branch:

```
BBL-PLA-BLACK   -> Black  #000000     (was "Blue")
BBL-PLA-IVORY   -> Ivory  #FFFFF0     (was "Orange")
BBL-PLA-BL      -> Blue   #0A2989     (two-letter codes still work, now anchored)
BBL-PLA-901234  -> colorName undefined, colorCode undefined   (no longer wipes the chosen colour)
[BBL]PCTG …     -> pctg, 250-270      (was pc, 260-290)
[BBL]PC …       -> pc,   260-290
```

**Material and manufacturer rewriting** is gone: `findMatchingMaterial("PLA (Silk)", ["PLA"])` now returns `undefined` instead of `PLA`, so an annotated entry is no longer silently rewritten on an unrelated save, and the dropdown keeps both.

**Seeding**, on a real `postgres:16-alpine`:

```
after migrate:           system_settings=1
seed --starter:          manufacturers=4  materials=4  colors=8  storage_locations=2
clear the lists, again:  "Sequence generator indicates starter data was previously seeded, skipping"
                         manufacturers=0
```

So the fresh-install regression is closed and Check 3 now actually works on Postgres — dropping `is_called` and testing `last_value` was the right call. The row-count assertions in both CI seed jobs mean this cannot regress silently again, which was the part that worried me most.

`npx playwright test` — 43 passed. `TEST_DIALECT=sqlite npx vitest run` — 625 passed, 1 skipped.

**One note on the two edited migrations**, since they are already on `main`. I checked the upgrade path: an install migrated on `main` (where the `INSERT` was silently dropped, so `system_settings` is empty) upgrades against this branch cleanly — exit 0, seven migrations recorded, nothing re-applied. Drizzle picks work by journal timestamp rather than by file hash, so amending an old file is inert for anyone who already ran it. Worth knowing that those existing SQLite installs still have no `system_settings` row and continue to rely on the lazy `onConflictDoNothing` creation in `getSystemSettings()`, which works. Only new installs get the row from the migration. Fine as it stands; just do not count on that row existing purely because the migration says so.

PR #40, opened 2026-09-14
feat: mobile-first community catalog search results with GTIN merging and overwrite prompt

Azumi0:
## Description

Enhances the community catalog search on the "Add Filament" modal to provide a mobile-first, informative experience:
1. **Mobile-First Rich Cards Component**: Replaces the single-line select dropdown with touch-friendly responsive cards (`CommunityCatalogSearchResults`). Cards display color swatch, manufacturer and filament name, color name, material badge, spool weight, diameter, spool/refill badges, extruder/bed temperatures, and GTIN chips.
2. **Accessible Sibling Button Architecture**: Card body and GTIN chips are structured as sibling buttons (no invalid nested interactive controls), meeting WCAG 2.2 touch target spacing guidelines (`min-h-[36px] min-w-[44px]` with spacing and `aria-label`). Badges inside buttons are rendered as `<span>` elements for strict HTML phrasing content compliance.
3. **Deduplication & Multi-GTIN Merging**:
   - Removes exact identical community catalog duplicates.
   - If two community items match on all physical specifications (`manufacturer`, `material`, `name`, `colorName`, `colorCode`, `density`, `diameter`, `weightGrams`, `spoolRefill`, `extruderTemp`, `bedTemp`) but differ in GTIN, they are merged into a single item with an array of unique `gtins`.
   - Shared deduplication utility in `shared/community-catalog-dedup.ts` used across both server and client.
   - Re-indexes all merged GTINs in `CommunityCatalogService.gtinMap` so scanning any merged barcode resolves to the canonical item.
4. **Barcode Overwrite Confirmation Dialog**:
   - If the modal's barcode/GTIN field is already filled and a community catalog item with a different GTIN is selected, an `AlertDialog` prompts the user to either keep their existing barcode or overwrite with the catalog GTIN.
   - Intelligently differentiates generic card selection (preserves scanned/entered barcode if it matches any of the item's merged GTINs) from explicit GTIN chip clicks (immediately updates to the clicked barcode without redundant prompts).
   - Dialogs are sequentially gated (`!similarManufacturerPrompt && !similarMaterialPrompt && overwriteBarcodePrompt`) to prevent modal stacking and focus traps.
   - Full translations in English, Polish, and German.

## Type of change

- [x] Code refactoring
- [x] Accessibility improvement

## How Has This Been Tested?

- **Unit & Component Tests** (`npm test`):
  - `tests/utils/community-catalog.test.ts`: Server-side deduplication, exact duplicate removal, multi-GTIN merging, and lookup indexing.
  - `tests/utils/community-catalog-helpers.test.ts`: Formatting helpers (`formatWeightGrams`, `formatDiameter`, `formatPrintTemps`, `deduplicateCommunityResults`).
  - `tests/components/community-catalog-search-results.test.tsx`: Card rendering, badges, color swatches, and multi-GTIN chips.
  - `tests/components/filament-modal-gtin-overwrite.test.tsx`: Overwrite prompt decision table, `resolveBarcodeUpdate` across single/multi-GTIN items, and locale key parity.
  - Full suite: **50/50 test files passed (646 passing tests, 9 skipped, 0 failures)**.
- **Type Checking & Linting**:
  - `npm run type-check`: 0 errors.
  - `npm run lint`: 0 warnings, 0 errors.
- **Production Build**:
  - `npm run build`: Vite build + esbuild (PG and SQLite) succeeded.
- **Playwright E2E Tests** (`npx playwright test tests/e2e/community-catalog-add-modal.spec.ts`):
  - **7/7 passed**, covering:
    1. OFD / SpoolmanDB source toggle and persistence.
    2. Searching community catalog and pre-filling all fields.
    3. Direct barcode entry and saving with barcode.
    4. Similarity prompt for materials.
    5. Similarity prompt for manufacturers.
    6. Barcode overwrite prompt (both "Keep" and "Overwrite with..." paths).
    7. Multi-GTIN chip selection and switching between merged barcodes.

## Checklist:


Azumi0:
This is a big painpoint for me, since I mostly add filaments from my mobile

the-luap (commented):
This is in good shape. I exercised the merging against the real Open Filament Database rather than fixtures, and the mechanics hold up. All ten checks are green. Three minor notes, none blocking.

## Verified against the live catalog

Syncing OFD on this branch and inspecting the result:

```
sync ofd                     22088 items in 532ms   (was 22354 before merging, so 266 collapsed)
items carrying >1 gtin       206
merged gtin lookups          401 checked, 0 unresolved
merging an already-merged set  items 22088 -> 22088, gtins 2693 -> 2693   (idempotent)
```

The idempotency matters more than it looks, because `extractUniqueGtins` reads both `gtin` and `gtins`, so a second pass over merged data is stable rather than compounding. Merging also runs from `setItems`/`setSourceItems` — at sync and disk load — not per search, which is the right place for an O(n) pass over twenty-odd thousand rows.

I also simulated a restart, since this is where a schema addition usually gets quietly dropped: a fresh service instance loading the snapshot from disk still has all 206 multi-GTIN items, and each merged GTIN still resolves. `sanitizeLoadedItems` carries the new field through.

The overwrite decision table behaves exactly as the description says, across every case I tried:

| entered | picked | explicit chip | result |
| --- | --- | --- | --- |
| nothing | 111 | no | update |
| 111 | 111 | no | none |
| 222 (same item) | 111 | no | none — keeps what was scanned |
| 999 | 111 | no | **prompt** |
| 222 (same item) | 111 | **yes** | update — no redundant prompt |
| 999 | 111 | **yes** | **prompt** |

And the component renders three buttons at nesting depth 1, so the sibling-button claim holds — no interactive control inside another — with both merged GTINs as chips and aria-labels present.

## Minor

1. **The phrasing-content claim is only half met.** Badges became `<span>`, but the card body button still wraps `<div>` layout elements (lines 81-82). `<button>`'s content model is phrasing content, so a `<div>` inside it is invalid HTML even though every browser renders it. Tailwind's `flex`/`block` work on a `<span>`, so this is a find-and-replace rather than a restructure — worth doing since the PR claims strict compliance.
2. **`gtins` is a new field on two API responses** — `GET /api/community-filaments/search` and `/gtin/:code` — and `docs/API.md` does not mention it. Anything consuming the API sees a shape change with nothing describing it.
3. `client/src/lib/community-catalog.ts` exports the same function under two names (`mergeCatalogItems` and `deduplicateCommunityResults`). One of them is enough; the tests use the alias, so renaming there would let the alias go.

Nothing here needs another round from my side — happy to see it merged with or without the notes addressed.

the-luap:
**Correcting my review above.** A second pass found two things I missed, one of them user-visible on every card. My own check hid the first: I rendered the component with a stub translator that returns its key, so the broken output looked like working output. Apologies.

## Blocking

**A. The spool-type badge prints a raw translation key** — `community-catalog-search-results.tsx:58-60`

```js
t("filamentsTable.spoolless") || t("filaments.spoolless") || "Refill"
```

There is no `filamentsTable` namespace in any locale; the real one is `filaments.*`, which `filament-modal.tsx:1228` already uses correctly. And `getTranslation` returns **the key itself** on a miss, which is truthy, so the `||` chain never reaches the working key or the literal. Against the real locale objects:

```
en  filamentsTable.spooled -> "filamentsTable.spooled"     filaments.spooled -> "Spooled (standard)"
de  filamentsTable.spooled -> "filamentsTable.spooled"     filaments.spooled -> "Spule (Standard)"
pl  filamentsTable.spooled -> "filamentsTable.spooled"     filaments.spooled -> "Na szpuli (standard)"
```

So every result card shows the literal text `filamentsTable.spooled` or `filamentsTable.spoolless` as its badge, in all three languages, and logs `Translation key not found:` on each render.

`tests/components/community-catalog-search-results.test.tsx:14` cannot catch it, because the mock `t` special-cases `filamentsTable.spoolless` and `filamentsTable.spooled` and returns "Refill"/"Spool" for them. The mock was written to match the implementation rather than the locales, so it encodes the bug. Worth asserting against the real locale object, or at least having the mock fall through to the key the way the real one does.

Fix is the two key names. The `||` fallbacks should go too — with this `t`, a fallback chain can never fire, so it reads as defensive but is inert.

**B. The overwrite dialog can ask you to replace a barcode with itself** — `filament-modal.tsx:321`

The explicit branch prompts whenever `current && !validGtins.includes(current)`, without first checking whether the two are the same string:

```
resolveBarcodeUpdate("111", "111", ["222"], true)
  -> { type: "prompt", currentBarcode: "111", incomingBarcode: "111" }
```

The user gets "The barcode field already contains 111. Overwrite it with 111?" over two buttons reading `Keep "111"` and `Overwrite with "111"`.

The realistic route is a scan. The field is set to the scanned code before the lookup, and the lookup passes that same code back as the chosen GTIN. It also fires on the leading-zero case the index is deliberately built to tolerate — a UPC-A scan of `684620401324` against a catalog entry stored as `0684620401324` resolves fine, then prompts, because the raw strings differ:

```
resolveBarcodeUpdate("684620401324", "684620401324", ["0684620401324"], true)
  -> prompt, both sides identical
```

An early `if (current === incoming) return { type: "none" }` covers both. Comparing normalised forms would also stop the leading-zero variant prompting at all.

**My decision table above was incomplete** — I tried eight cases and none had `current === incoming` with an explicit chip, which is exactly the gap. The table is right for the cases it covers; it just does not cover this one.

## Worth a look

- **Two dialogs can mount together.** The overwrite prompt is gated on the two similarity prompts but not on `variantCandidates`, and the scan path can set both, giving competing focus traps.
- **"Keep existing barcode" leaves the form mixed.** Every other field is overwritten unconditionally when a result is picked, so keeping the old barcode leaves filament B's details carrying filament A's barcode, and it saves that way.
- **`formatPrintTemps(null, 60)` returns null**, so an entry with a bed temperature but no extruder temperature shows no temperature badge at all. SpoolmanDB has such entries.
- **Two untranslated strings** in a PR that otherwise adds all three languages: `"Bed"` in the temperature string, and `aria-label={\`Select GTIN ${gtin}\`}`.

The server-side merging, indexing, disk round-trip and idempotency I verified earlier all still stand — those were the parts I checked against the live catalog rather than through a stub.

Azumi0:
Addressed all feedback points:

1. **Spool-type badge translation keys**:
   - Switched to `filaments.spoolless` and `filaments.spooled` in `client/src/components/community-catalog-search-results.tsx`.
   - Dropped the dead `||` fallback chains.
   - Updated component tests in `tests/components/community-catalog-search-results.test.tsx` to assert against real locale translations.

2. **Self-replacement & leading-zero barcode prompts**:
   - `resolveBarcodeUpdate` now returns `{ type: "none" }` early when `current === incoming` or when normalized forms match (`normalizeGtin(current) === normalizeGtin(incoming)`).
   - `itemGtins` membership checks now compare normalized forms, avoiding false prompts on UPC-A vs EAN-13 scans with leading-zero differences.
   - Added unit test coverage for self-replacement and leading-zero variants in `tests/components/filament-modal-gtin-overwrite.test.tsx`.

3. **Button phrasing content**:
   - Replaced layout `<div>` elements inside the card `<button>` with `<span>` elements styled with Tailwind `block` and `flex` classes.

4. **Dialog gating**:
   - Gated `overwriteBarcodePrompt` on `!variantCandidates` to prevent simultaneous mounting and competing focus traps when scanning barcodes with multiple variant candidates.

5. **Bed-only temperatures & missing translations**:
   - Updated `formatPrintTemps` to support bed-only profiles (`Bed ${bedTemp}°C`), and aligned `handleUseCommunityResult` in `filament-modal.tsx` to use `formatPrintTemps` consistently.
   - Localized `"Bed"` (`filaments.bed`) and the GTIN chip aria-labels (`scanner.selectGtin`) across English, German, and Polish.

6. **Helper export alias**:
   - Removed the `deduplicateCommunityResults` alias in `client/src/lib/community-catalog.ts`, standardizing on `mergeCatalogItems`.

7. **API documentation**:
   - Documented the `gtins` field on `GET /api/community-filaments/search` and documented `GET /api/community-filaments/gtin/:code` in `docs/API.md`.

8. **ADR 0011 (Barcode preservation)**:
   - Added `docs/adr/0011-community-catalog-barcode-overwrite-preservation.md` formalizing the workflow: scanning an unlisted physical barcode, searching the catalog to populate filament specifications, and intentionally retaining the scanned barcode on the spool rather than overwriting it with the catalog GTIN.

All 50 test suites (651 tests), linting, type-checking, and Playwright E2E specs pass.

the-luap (commented):
Reviewed `1fc3cff` and `510a790`. Both blockers are fixed and every smaller point is addressed. All ten checks are green. Ready to merge from my side.

This time I checked the rendering with the **real** translator against the real locale objects, since a stub is what hid the badge bug from me in the first place:

```
filaments.spooled    -> ["Spooled (standard)", "Spule (Standard)", "Na szpuli (standard)"]
filaments.spoolless  -> ["Spoolless", "Spulenlos", "Bez szpuli"]
filaments.bed        -> ["Bed", "Bett", "Stół"]
scanner.selectGtin   -> ["Select GTIN {{gtin}}", "GTIN {{gtin}} auswählen", "Wybierz GTIN {{gtin}}"]

rendered in en / de / pl:  raw keys leaked = none
unresolved {{placeholders}} = none
aria-labels = ["GTIN 111 auswählen", "GTIN 222 auswählen"]
```

So the badges read properly in all three languages and the chip labels interpolate. Updating the component test to assert against the real locales rather than a hand-written mock is the part that stops this recurring.

The barcode prompt now stays out of the way exactly where it should and still appears where it matters:

```
identical codes           -> none
UPC-A vs stored EAN-13    -> none      (684620401324 against 0684620401324)
genuinely different code  -> prompt    (999 against 111)
```

Also verified: no `<div>` remains inside the card button and nesting depth is still 1, so the phrasing-content claim now holds in full; `formatPrintTemps(null, 60)` gives `"Bed 60°C"` instead of nothing; the `gtins` field is documented on both endpoints; and the duplicate export alias is gone.

On ADR 0011 — documenting "keep the scanned barcode" as the intended workflow rather than changing the behaviour is the right call, and better than what I suggested. Scanning a spool whose barcode the catalog does not list, then using the catalog only for its specifications, is a real thing people will do, and the barcode is the one field that is genuinely theirs. Worth noting the ADR is what makes it a decision rather than an inconsistency.

Nothing outstanding.

PR #41, opened 2026-09-15
feat: resolve scanned barcodes against personal collection first (Priority 0)

Azumi0:
## Description

When scanning a barcode to add or look up a filament spool, Filadex previously bypassed the user's existing inventory and queried external community catalogs (OpenFilamentDatabase / SpoolmanDB) directly. When users had already registered a spool with that barcode, rescanning another spool of the same batch or re-adding an identical spool required manual re-entry or resulted in external catalog queries that might not reflect user-tailored specifications.

This PR establishes **Priority 0 (Personal Collection)** resolution in the barcode lookup hierarchy:
1. **Priority 0 (Personal Collection)**: Exact match or leading-zero normalized barcode match across existing spools.
2. **Priority 1 (Community Catalogs)**: OpenFilamentDatabase / SpoolmanDB GTIN lookup.
3. **Priority 2 (Manual Entry / Barcode-Only Fallback)**: Preserves the scanned barcode in the form for manual completion.

### Key Behaviors & Architecture
- **Specifications vs Instance Separation**: When cloning from an existing collection match, only physical product specifications (`name`, `manufacturer`, `material`, `colorName`, `colorCode`, `diameter`, `printTemp`, `totalWeight`, `spoolType`, `barcode`) are copied. Instance-specific metadata (`storageLocation`, `remainingPercentage`, `purchasePrice`, `purchaseDate`, `customFieldValues`, `usageLog`) starts with empty/clean defaults for a new spool.
- **Leading-Zero Normalization**: Handles UPC-A / EAN-13 padding (e.g. `0123...` matches `123...`) and case-insensitive alphanumeric barcodes.
- **Identical vs Conflicting Matches**:
  - If multiple existing spools share identical product specifications (or the same `filamentTypeId`), the latest spool is resolved cleanly.
  - If multiple spools match the barcode with conflicting specifications (e.g. different materials or colors), an interactive candidate picker dialog is displayed allowing the user to select the desired spool profile.
- **Escape Hatch**: A toast action `"Search Community Catalog instead"` allows on-demand querying of external catalogs with visual feedback if not found.
- **Collection View Actions**: Scanning on the main screen prompts the user with `"View in Collection"` (to inspect the existing spool) or `"Add Another Spool"` (to open the Add modal pre-filled with the matched spool's product specifications), with explicit Cancel support.
- **Self-Exclusion**: When editing an existing spool, that spool's ID is excluded from collection lookup so it does not match itself.
- **Architecture & Domain Documentation**: Added [ADR 0012](docs/adr/0012-priority-0-barcode-resolution-from-personal-collection.md) and updated `CONTEXT.md` with canonical definitions.

## Type of change
- [x] Code refactoring

## How Has This Been Tested?
- **Unit & Component Tests (`npm test`)**:
  - `tests/lib/collection-lookup.test.ts`: Barcode normalization (leading zeros, whitespace, alphanumeric case-insensitivity), single match, conflict handling, and product spec extraction.
  - `tests/lib/collection-scanner-translations.test.ts`: Translation key existence and interpolation verification across `en`, `de`, and `pl`.
  - `tests/components/filament-modal-priority-0.test.tsx`: Render tests for modal in Add mode with copied spool draft, candidate dialog, and self-exclusion when editing.
  - Full suite: 53/53 test files passed (665 passing tests, 9 skipped, 0 failures).
- **Type Checking**:
  - `npm run check` (tsc): 0 errors.
  - `npm run check:sqlite`: 0 errors.
- **i18n Parity**:
  - Complete translations added in English (`en.ts`), German (`de.ts`), and Polish (`pl.ts`).

## Checklist:


the-luap (commented):
Solid work, and the separation at the heart of it is the right idea: a barcode identifies a *product*, but a spool is an *instance*, so cloning specifications while resetting remaining percentage, price, location and custom fields is exactly the distinction that makes rescanning useful. All ten checks are green. Nothing blocking; three notes, one of which is not this PR's doing.

## Verified

I exercised `collection-lookup.ts` directly rather than reading it:

```
norm("0123")=123   norm("000")=0   norm("")=""   norm("  0123  ")=123   norm("abc123")=abc123
scan 123456789012 / 0123456789012 / 00123456789012  against a spool stored as 0123456789012 -> single
scan "nope" -> not_found        scan "" -> not_found
conflict with PLA, PETG, PLA    -> conflict, 2 unique candidates from 3 matches
extractProductSpecsFromSpool    -> leaked instance fields: []   (id, remainingPercentage,
                                   purchasePrice, purchaseDate, storageLocation,
                                   customFieldValues all absent)
```

The leading-zero handling is right in both directions, the all-zeros case collapses to `"0"` rather than empty, and alphanumeric codes keep their case for the normalised form while matching case-insensitively.

`applyCollectionSpoolData` sets fields one at a time rather than spreading the matched spool, so there is no path for instance data to reach the form even if the extractor were widened later. Self-exclusion is correct: `isEditing && filament?.id ? raw.filter(f => f.id !== filament.id) : raw`.

## Notes

**1. No browser coverage for a feature that is entirely interaction.** Every file added under `tests/` is a unit or `renderToString` test, and the new component test has to mock Dialog and Select to make children render at all — which is the tell. `CLAUDE.md` §5 says plainly that `renderToString` runs no effects and to reach for Playwright for "effect ordering and remounts, navigation, cookies". This feature is a scan, then a toast with two actions, then possibly a candidate dialog, then a prefilled modal. The four PRs before this one each added an e2e spec; this is the first that does not, and it is arguably the most dialog-driven of them. A spec covering scan → "Add Another Spool" → assert the form carries the specs and *not* the remaining percentage would pin the thing the PR is actually for.

**2. Weight and spool type are copied but never compared.** `haveSameProductSpecs` compares manufacturer, material, colour, code, diameter and print temp; `extractProductSpecsFromSpool` also copies `totalWeight` and `spoolType`. So two spools sharing a barcode but differing only in size resolve as `single` and silently take the newest:

```
1000g spooled (id 1) and 250g spoolless (id 2), same barcode
  -> haveSameProductSpecs = true, result = single, chosen id 2 (250g, spoolless)
```

Different sizes usually carry different barcodes, so this is an edge, but the two functions disagree about what counts as the same product. Either include them in the comparison, or leave them out of the copy.

## Not this PR, but worth knowing

`t('settings.materials.invalidDensity')` at `filament-modal.tsx:196-197` resolves to nothing in all three locales — the key does not exist — so the density validation message reads as the literal key. This is on `main` already (it arrived with the density validation in #36) and these lines are untouched here, so it is not yours to fix in this PR. Flagging it because it is the **second** instance of the exact pattern we hit on #40: a `t()` call for a missing key with a `|| 'fallback'` that can never fire, because `getTranslation` returns the key itself and a key is truthy.

Your `collection-scanner-translations.test.ts` is the right instinct, but it checks a hand-written `requiredKeys` list, so it only covers keys somebody remembered to add. A test that greps the source for `t('...')` literals and asserts each one resolves in en/de/pl would have caught both #40's badge and this density message, and would keep catching them. Cheap, and it would have made my last two reviews unnecessary.

the-luap:
**Correcting my review above — "nothing blocking" was too generous.** A second pass found three more, all verified on the branch. The first is the one I would fix before merging.

## Should fix before merge

**A. Priority 0 is not gated on Add mode, so scanning while editing rewrites the edited spool's identity** — `filament-modal.tsx:1245`

`handleQRCodeScanned` runs `resolveCollectionBarcode` unconditionally. Open **Edit** on spool A, scan a barcode belonging to spool B, and `applyCollectionSpoolData(B)` overwrites A's `name`, `manufacturer`, `material`, `colorName`, `colorCode`, `diameter`, `printTemp`, `totalWeight` and `spoolType` — with a toast, no confirmation, and no way back other than cancelling the edit. A user scanning inside an edit almost certainly means "set this spool's barcode", not "replace what this spool is".

ADR 0012 already says the right thing: it scopes Priority 0 to the *Add Spool modal*. The code does not. Gating on `!isEditing` makes them agree. Note this is separate from the self-exclusion filter, which is correct — that stops A matching *itself*; nothing stops A being overwritten by B.

**B. `collectionCandidates` is never reset, and a stale value permanently disables the barcode-overwrite prompt** — `filament-modal.tsx:777`, `2555`

`resetToDefaults()` clears `variantCandidates` and `variantCandidateBarcode` but not `collectionCandidates` / `collectionCandidateBarcode`. The picker renders outside the `isOpen` gate as `<Dialog open={true}>`, and `home.tsx:704` mounts the modal unconditionally, so it is never unmounted. Close the modal while the picker is open and the value survives — which then trips this guard for the rest of the page's life:

```jsx
{!collectionCandidates && !variantCandidates && … && overwriteBarcodePrompt && (
```

So the barcode-overwrite confirmation that #40 was all about stops appearing entirely, silently, until a reload. Adding the two setters to `resetToDefaults` covers it.

**C. A failed catalog lookup reports "not recognized"** — `filament-modal.tsx:1180`, `1283`

`queryCommunityCatalogGtin` returns `false` both for a genuine 404 and, after already showing a destructive toast, for a 500 or network failure. The caller reads `false` as "not found" and falls through, so a server error produces two destructive toasts in a row: the real error, then `scanner.notFoundAllSources` — "Barcode X not recognized (searched Filadex QR, Bambu Lab, and Open Filament Database)", which is untrue, nothing was searched. That is a regression of the error differentiation added in #36; `home.tsx` still gets it right with its `lookupFailed` flag. Returning `"found" | "not_found" | "error"` would restore the distinction.

## Sharpening the weight/spool-type point from my review

The conflict picker displays exactly `candidate.totalWeight` and `candidate.spoolType` as the attributes distinguishing one candidate from another (`filament-modal.tsx:1330-1342`, `home.tsx:869-881`) — and those are precisely the two fields `haveSameProductSpecs` ignores. So when they are the only difference the result is `single` and the picker never opens; when the picker does open they are guaranteed identical. The columns can never show a difference. Either compare them or stop displaying them.

## Minor

- The "Search Community Catalog instead" escape hatch is steered by the data it exists to escape: it reads `diameter`, `totalWeight` and `spoolType` from the form, which `applyCollectionSpoolData` has just overwritten with the matched spool's values, and sends them as the hint the server uses to pick a primary candidate. Harmless when the collection entry is the same variant, self-defeating when it is not.
- ADR 0012 lists `density` among the copied specifications; `extractProductSpecsFromSpool` does not extract it and `applyCollectionSpoolData` never sets it. One of the two should change.
- `scanner.selectSpool` is added to all three locales and asserted by the new translation test, but never used in code.

Everything I verified in the first review still stands — the normalisation, matching, conflict grouping, absence of instance-field leakage and self-exclusion are all correct. These are the paths around it.

Azumi0:
Addressed the review feedback in commits `a065531` and `e00a6c6`:

### 1. Edit mode Priority 0 gating with confirmation (Issue A)
* When editing an existing spool (`isEditing`), scanning a barcode matching another spool in the collection updates the barcode field immediately and presents a confirmation dialog (`overwriteSpecsTitle` / `overwriteSpecsDescription`).
* The user can select **"Update Barcode Only"** (leaving the current spool's specifications intact) or **"Overwrite Specifications"** (applying the matched spool's specs).
* Resolving a conflict candidate while editing similarly routes through this confirmation prompt.

### 2. Candidate state reset (Issue B)
* `resetToDefaults()` in `filament-modal.tsx` now clears `collectionCandidates`, `collectionCandidateBarcode`, and `overwriteSpecsPrompt`.
* The candidate dialog's `onOpenChange` and explicit Close button now also synchronize clearing both `collectionCandidates` and `collectionCandidateBarcode`.

### 3. Error differentiation on community catalog lookup (Issue C)
* Refactored `queryCommunityCatalogGtin` to return `"found" | "not_found" | "error"`, matching `home.tsx`'s error handling.
* Non-404 errors show only the destructive error toast (`scanner.lookupError`) and do not fall through to the duplicate `scanner.notFoundAllSources` toast.

### 4. Weight and spool type comparison (Note 2)
* Updated `haveSameProductSpecs` in `collection-lookup.ts` to compare `totalWeight` and `spoolType` alongside material specifications.
* Two spools sharing a barcode that differ in size (e.g. 1000g vs 250g) or packaging (spooled vs refill) now produce a `conflict` result and open the candidate picker. Added unit test in `tests/lib/collection-lookup.test.ts`.

### 5. Unbiased escape hatch lookup (Minor 1)
* Added `ignoreFormHints` option to `queryCommunityCatalogGtin`.
* Clicking "Search Community Catalog instead" now passes `ignoreFormHints: true`, omitting the matched spool's newly copied form attributes so the community search is unbiased.

### 6. ADR 0012 alignment (Minor 2)
* Removed `density` from the copied specifications list in ADR 0012, aligning with the schema where density is a material-level attribute rather than a spool column.
* Documented the Edit mode confirmation prompt and unbiased escape hatch parameters.

### 7. Translation cleanup (Minor 3)
* Removed the unused `scanner.selectSpool` key from `en.ts`, `de.ts`, `pl.ts` and updated `collection-scanner-translations.test.ts`.
* Added keys for the Edit mode confirmation dialog across all three locales.

### 8. Missing `invalidDensity` key & automated translation test (Review note)
* Added `settings.materials.invalidDensity` across `en.ts`, `de.ts`, and `pl.ts`.
* Added `tests/i18n/source-translation-keys.test.ts`, which scans all `.ts`/`.tsx` files in `client/src` for static `t('...')` literals and asserts each key resolves to a non-empty string in all three locales.

### 9. Playwright browser coverage (Note 1)
* Added a `filadex:scan` custom event listener in `QRScanner` to enable headless scan simulation.
* Added `tests/e2e/priority-0-barcode-lookup.spec.ts` exercising:
  1. Home scanner -> match existing collection spool -> prompt "Add Another Spool" -> asserts product specs are prefilled and remaining percentage resets to 100% (not the seeded spool's 35%).
  2. Edit mode scanner -> scan another collection spool's barcode -> asserts confirmation prompt appears, and choosing "Update Barcode Only" updates the barcode while preserving identity.
  3. Edit mode scanner -> choosing "Overwrite Specifications" updates the specifications while preserving spool instance attributes.

the-luap (commented):
Reviewed `a065531` and `e00a6c6`. Everything from both rounds is addressed, and I checked each one on the branch rather than taking the summary's word for it. All ten checks are green, 678 unit tests and 48 browser specs pass locally. Ready to merge from my side.

**Edit mode now asks.** `handleQRCodeScanned` sets the barcode and then, when `isEditing`, raises `overwriteSpecsPrompt` and returns instead of applying the specs. That is the right split: scanning inside an edit sets the barcode, which is what the user meant, and touching the spool's identity needs a yes.

**The comparison and the picker agree now:**

```
1000g spooled vs 250g spoolless, same barcode -> sameSpecs=false, conflict, 2 candidates
genuinely identical                           -> single, newest chosen
```

So the two columns the picker displays can finally differ in the case it opens for.

**Also verified:** `queryCommunityCatalogGtin` returns `"found" | "not_found" | "error"` and the caller only reports "not recognized" on `not_found`, so a server failure no longer produces two contradictory toasts; `resetToDefaults` clears `collectionCandidates`, `collectionCandidateBarcode` and `overwriteSpecsPrompt`, closing the leak that permanently suppressed the overwrite confirmation; the escape hatch passes `ignoreFormHints: true`; ADR 0012 no longer claims `density`; and the unused `scanner.selectSpool` is gone.

**The translation test is the part I am most pleased about, and I checked that it actually bites.** Appending `t("totally.missing.key")` to a source file turns it red in all three locales with the key named:

```
The following translation keys are referenced via t('...') in client/src but missing or empty in en.ts:
totally.missing.key
```

The `\bt\(` boundary also avoids the false positives my own crude grep hit (it does not match `get(`). That single test would have caught the #40 badge, the density message and anything of the same shape in future — worth more than the two fixes it shipped with. `settings.materials.invalidDensity` is now present in all three locales, which closes #42's specific half.

The three new browser specs cover exactly the paths that needed a real browser, including asserting the remaining percentage resets to 100% rather than inheriting the matched spool's 35%.

## One thing to be aware of, not a blocker

`qr-scanner.tsx:544` registers a `filadex:scan` window listener that is not gated by environment, so it ships in production builds. I do not think it is exploitable — the listener only exists while the scanner dialog is mounted, and anything able to dispatch the event can already type into the form or call the API with the user's session — so it grants no capability an attacker lacks. But it is a test affordance in shipped code, and the obvious guard is unavailable to you because the e2e suite deliberately runs the production bundle, so `import.meta.env.DEV` would switch it off exactly where it is needed.

If you want it gone from real builds, a build-time flag that `playwright.config.ts` sets for its own build would do it. Otherwise a comment at the listener saying it is deliberate, and why the obvious gate does not work, would stop the next reader filing it as a mistake.

Azumi0:
Good catch on the test affordance. Gating it out of real builds is definitely the right call, especially since test affordances tend to linger once accepted.

This is now resolved in `6077327` and `a30983f`:

1. **Dead-Code Elimination via Build-Time Flag**:
   - Extracted the window listener into a dedicated `<E2EScanAffordance />` helper component, conditionally mounted in `QRScanner`:
     ```tsx
     {import.meta.env.VITE_TEST_AFFORDANCES === "true" && (
       <E2EScanAffordance onScan={handleScanSuccess} />
     )}
     ```
   - In standard release builds (`npm run build`, Docker, Synology), `VITE_TEST_AFFORDANCES` is unset. Vite statically inlines `undefined === "true"` to `false`, and Rollup prunes the JSX and tree-shakes `<E2EScanAffordance>` completely. The output bundle contains zero occurrences of `"filadex:scan"`.

2. **Dedicated E2E Build Script & Playwright Diagnostics**:
   - Added `"build:e2e": "VITE_TEST_AFFORDANCES=true npm run build"` in `package.json`.
   - Updated `.github/workflows/test.yml` to use `npm run build:e2e` for the `e2e` job.
   - Updated `playwright.config.ts` with a pre-flight check: if client assets exist but lack test affordances (e.g. from a prior standard `npm run build`), Playwright throws an immediate error explaining that `npm run build:e2e` is required, avoiding confusing timeouts.

3. **CI Production Build & Bundle Leak Verification**:
   - Added `scripts/verify-bundle-affordances.ts` (`npm run verify:bundle`), which inspects `dist/public/assets/*.js` and asserts that neither `filadex:scan` nor `E2EScanAffordance` exist in the bundle.
   - Added a dedicated `build` job in `.github/workflows/test.yml` running `npm run build && npm run verify:bundle` on every push and PR to verify standard release builds don't break and don't leak test code.

4. **Documentation**:
   - Recorded the rationale and constraints in [ADR 0013: Gating Test Affordances from Production Bundles via Build-Time Environment Flags](docs/adr/0013-test-affordance-build-gating.md).
   - Updated `docs/TESTING_GUIDE.md` to cover `npm run verify:bundle` and the `build:e2e` requirement.

the-luap (commented):
Reviewed `6077327` and `a30983f`. This went further than I suggested and the result is better. Verified on the branch:

```
standard `npm run build`   filadex:scan = 0   E2EScanAffordance = 0
npm run verify:bundle      PASS: Standard production bundle is clean

npm run build:e2e          filadex:scan = 2   (present, as the specs need)
npm run verify:bundle      LEAK: Found test affordance keyword "filadex:scan" in
                           dist/public/assets/index-B5ismg4K.js
                           FAIL
```

That second half is the part that matters. A guard which always passes is worse than none, and this one genuinely fails on a bundle that carries the affordance, naming the file. So the `build` job now proves the claim on every push rather than restating it.

The Playwright pre-flight also does what it says — running the specs against a standard build stops immediately with `dist/public/assets was built without test affordances ... Run npm run build:e2e`, instead of timing out on a scan that never arrives.

48 browser specs, 678 unit tests, both type-check targets and lint all pass locally. ADR 0013 records why the obvious `import.meta.env.DEV` gate was unavailable, which is the bit a future reader would otherwise have to rediscover.

Merging.

PR #43, opened 2026-09-16
fix(i18n): drop dead || translation fallbacks and add guard test (#42)

### LINKED ISSUE #42: Missing translation keys render as raw keys, and the || fallback pattern can never catch them
Two instances of the same bug have now reached `main`, so this is about the pattern as much as the current occurrence.

## The pattern

`getTranslation` (`client/src/i18n/index.ts`) returns **the key itself** when a key is missing. A key is a non-empty string, so it is truthy, so this idiom cannot work:

```ts
t('some.missing.key') || t('some.other.key') || 'Readable fallback'
```

The first call returns `"some.missing.key"`, the chain short-circuits there, and the user sees the raw key. The fallback is dead code that reads like a safety net.

## Current occurrence

`client/src/components/filament-modal.tsx:196-197`

```ts
.regex(/^\d+(\.\d+)?$/, t('settings.materials.invalidDensity') || 'Density must be a positive number')
.refine((v) => Number(v) > 0, t('settings.materials.invalidDensity') || 'Density must be a positive number'),
```

`settings.materials.invalidDensity` is in none of `en.ts`, `de.ts`, `pl.ts` (verified against the real locale objects). So entering a bad density shows the validation message `settings.materials.invalidDensity` instead of a sentence, in all three languages, and logs `Translation key not found:` each time.

Arrived with the density validation in #36.

## Previous occurrence

The community catalog spool-type badge in #40 used a `filamentsTable.*` namespace that does not exist, with the same `||` chain. Every search result card showed the literal key as its badge until it was fixed. Its component test could not catch it because the mock `t` special-cased those keys and returned real words.

## Suggested fix

1. Add the missing key to the three locales, and drop the `||` fallbacks wherever they appear — with this `t`, they are always dead.
2. Add a test that finds the problem class rather than one instance: scan `client/src` for `t('...')` / `t("...")` string literals and assert each resolves in `en`, `de` and `pl`.

`tests/lib/collection-scanner-translations.test.ts` (from #41) is close, but it checks a hand-maintained `requiredKeys` array, so it only covers keys somebody remembered to list. Deriving the list from the source is what makes it durable. A crude grep over the three relevant files already finds this one, so it does not need to be sophisticated — it needs to be automatic.

Caveat for whoever writes it: dynamic keys (template literals, variables) cannot be resolved statically and should be skipped rather than reported, or the test will be noisy.

Azumi0:
## Description

Closes #42.

`getTranslation` in `client/src/i18n/index.ts` returns the key string itself when a key is missing. Because a non-empty string is truthy and non-null, fallback chains like `t('missing.key') || 'Fallback'` or `t('missing.key') ?? 'Fallback'` short-circuit on the key string, leaving the fallback unreachable dead code while displaying the raw key.

This PR addresses the issue by:
1. **Removing dead fallbacks across `client/src`**:
   - `client/src/components/filament-modal.tsx`: density validation schema (`regex` and `refine`), material preset labels, color preset labels, community catalog search and scan error toasts, candidate dialog spool type indicators and close buttons, "Other" manufacturer select item, and usage note/history labels.
   - `client/src/pages/home.tsx`: error toasts, candidate dialog spool type indicators, and close buttons.
   - `client/src/components/theme-selector.tsx`: theme preset colors and theme update toast description.
   - `client/src/components/community-catalog-search-results.tsx`: loading label.
   - `client/src/components/my-requests-modal.tsx` & `client/src/components/settings/settings-catalog-requests.tsx`: catalog request entity type titles.
2. **Adding an automated guard test in `tests/i18n/source-translation-keys.test.ts`**:
   - Added `it("does not use dead || or ?? fallbacks after t(...) in client/src")`: scans client source files for any `t(...)` call followed by `||` or `??` (including inside nested calls). This guarantees CI fails if a dead fallback is ever reintroduced.
   - Added `it.each(locales)("resolves settings.${entityType}s.title in %s for all catalogRequestEntityTypes")`: ensures all dynamic catalog request entity type titles resolve across `en`, `de`, and `pl`.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Code refactoring

## How Has This Been Tested?

- **Automated translation tests**:
  - `npx vitest run tests/i18n/source-translation-keys.test.ts` (8 passed):
    - Confirmed regression guard catches `t(...) ||` and `t(...) ??` occurrences.
    - Verified all static `t('...')` keys in `client/src` resolve in `en`, `de`, and `pl`.
    - Verified all `catalogRequestEntityTypes` resolve in `en`, `de`, and `pl`.
- **Targeted i18n & component tests**:
  - `tests/i18n/resolve-client-language.test.ts` (20 passed)
  - `tests/components/community-catalog-search-results.test.tsx` (5 passed)
  - `tests/i18n/general-settings-keys.test.ts` (2 passed)
  - `tests/lib/collection-scanner-translations.test.ts` (2 passed)
  - `tests/i18n/locale-parity.test.ts` (2 passed)
- **Full test suite**:
  - `npm run test`: 54 test files, 671 tests passed, 0 failures.
- **Type checking and linting**:
  - `npm run check` (`tsc`): 0 errors.
  - `npm run check:sqlite`: 0 errors.
  - `npm run lint` (`eslint`): 0 warnings, 0 errors.

## Checklist:


the-luap (approved):
Thanks, this is a clean fix and it closes the rest of #42. The missing `invalidDensity` key and the static-key completeness test already landed with #41, so this PR covers the part that was left: removing the dead fallbacks and adding a guard that catches the whole class of bug.

**What I checked**
- Every removed fallback really was dead code. `getTranslation` returns the key when a translation is missing, and the existing completeness test rejects empty values. So none of the removed `||`/`??` branches could ever run, and removing them doesn't change behavior. All the affected keys resolve in `en`, `de` and `pl`.
- The two dynamic `settings.${entityType}s.title` sites used to fall back to `request.entityType`. That fallback was dead too, and the new `catalogRequestEntityTypes` test replaces it with a real check.
- **The guard catches regressions:** I reverted only the `client/src` changes and kept the new test. `does not use dead || or ?? fallbacks` then fails and lists the original sites (`community-catalog-search-results.tsx:34`, `filament-modal.tsx:35`, …).
- **Blind spots the guard ignores correctly:** it doesn't flag `x || t('...')` (a real fallback to a translation). It does flag `i18n.t(...) ||`, nested template literals, and ternary keys.
- The `tests/i18n` suite passes locally (32 tests), and CI is green.

**Non-blocking notes**
1. The scanner has a few known gaps. It misses `(t('a')) || 'x'` and `t('a') as string || 'x'`, and it can lose track of quotes if a comment with an apostrophe (`// don't`) appears inside a `t(` argument list. None of these occur in the code today, and the issue asked for something automatic rather than exhaustive, so I'm fine leaving them. A short comment above `findDeadTranslationFallbacks` saying it is a heuristic, not a parser, would help the next reader.
2. Out of scope, but noticed: `filament-modal.tsx:1230` (`'Failed to fetch filament from label'`) and `home.tsx:389` (`'Failed to open filament'`) still use hardcoded English after `err?.message ||`. That fallback is reachable, so it isn't the same bug. It's just untranslated. Worth a follow-up if you want the scanner error paths fully localized.
3. Nit: the test file now ends with an extra blank line.

Approving.

Azumi0:
Addressed the non-blocking notes in commit f2132e7:

1. **Fallback Scanner Heuristic**: Added a docstring comment above `findDeadTranslationFallbacks` in `tests/i18n/source-translation-keys.test.ts` documenting that it is a fast heuristic character scanner rather than a full AST parser, explicitly describing its scope and known boundaries.
2. **Scanner Error Localization**: Added `fetchFromLabelError` and `openFilamentError` to `en.ts`, `de.ts`, and `pl.ts`. Updated `filament-modal.tsx` and `home.tsx` to use `err?.message || t('scanner.fetchFromLabelError')` and `err?.message || t('scanner.openFilamentError')`.
3. **Cleaned up formatting**: Removed the extra trailing newline at the end of `tests/i18n/source-translation-keys.test.ts`.

the-luap:
Checked f2132e7. It covers all three notes:

- `scanner.fetchFromLabelError` and `scanner.openFilamentError` are in `en`, `de` and `pl`, and they're placed next to `lookupError`. Both toasts now use them after `err?.message ||`, which is a real fallback, so the guard correctly leaves it alone. The static-key completeness test covers the new keys.
- The comment on the scanner describes what it catches and names the known gaps.
- The trailing newline is gone.

CI is green on the new head. My approval still stands. Thanks!

PR #44, opened 2026-09-16
feat(scanner): redesign mobile-first scanner modals and candidate cards

Azumi0:
## Summary

Addresses mobile usability and layout issues in the barcode scanner workflow where visual components overflowed horizontally and transient toasts with interactive buttons expired before users could make a decision.

This PR focuses strictly on the **Mobile-First Scanner Modals & Candidate Cards** after the clean separation of PR #45 (datepicker timezone), PR #46 (API error contract), PR #47 (toast lifecycle & accessibility), and PR #48 (mobile guidelines).

---

### Key Changes

1. **Modal-on-Modal Conscious Decision**:
   - Replaced auto-dismissing toasts in `FilamentModal` with an overlaying `AlertDialog` (`collectionMatchPrompt`) when a scanned barcode matches an existing spool in the user's collection.
   - Displays a static preview `CollectionSpoolCard` and requires a deliberate user choice: **"Use Collection Specs"**, **"Search Community Catalog instead"**, or **"Cancel"**.
   - Buttons stack vertically on mobile screens within thumb reach and wrap gracefully on desktop (`max-w-xl sm:flex-wrap`) preventing label truncation.

2. **Canonical Mobile-First Spool Card (`CollectionSpoolCard`)**:
   - Consolidates three disparate, unstyled card implementations across `home.tsx` and `filament-modal.tsx` into a single canonical component adopting the PR #40 visual standard.
   - **High-contrast color swatch**: 16px circle with `border border-black/15 dark:border-white/20 shadow-sm mt-0.5` ensuring visibility of light colors in light mode and dark colors in dark mode.
   - **Material badge**: Prominent secondary badge (`bg-neutral-100 dark:bg-neutral-700/60`).
   - **Metadata pills**: Outlined badges for spool type (with emerald accent for spoolless), weight, diameter, and print temperature with wrapped alignment (`pl-[26px]`).
   - **Robust text wrapping**: `min-w-0 break-words` prevents title or manufacturer overflows.
   - **Accessibility & Touch**: Minimum 44px touch targets (`min-h-[44px]`) and `focus-visible:ring-2 focus-visible:ring-inset`.
   - Supports both interactive selection and static preview (`interactive={false}`).

3. **Responsive Dialog Constraints**:
   - Base `DialogContent` and `AlertDialogContent` styled with `w-[calc(100vw-2rem)] sm:w-full max-w-lg` (and `max-w-xl` for multi-action decision dialogs) and `max-h-[95vh] sm:max-h-[90vh] overflow-y-auto` to eliminate horizontal viewport clipping on 320px–390px mobile screens.
   - Standardized dialog footers (`flex flex-col sm:flex-row-reverse gap-2`) per mobile-first guidelines.

4. **Test Suite Hygiene & Separation**:
   - Toast presentation and touch-accessibility test relocated to its dedicated suite `tests/e2e/toast-auto-dismiss.spec.ts`.
   - `tests/e2e/priority-0-mobile-decision.spec.ts` strictly focuses on scanner decision dialogs and candidate cards.

---

### 1. Candidate Card: Before vs. After

Replaces 3 disparate flat button cards with the canonical mobile-first `CollectionSpoolCard`.

| Light Theme | Dark Theme |
|---|---|
| Candidate Card Before vs After - Light | Candidate Card Before vs After - Dark |

---

### 2. Collection-Match Conscious Decision Dialog

Overlaying `AlertDialog` replacing the auto-dismissing toast when a scanned barcode matches an existing collection spool.

| Viewport / Theme | Light Theme | Dark Theme |
|---|---|---|
| **Mobile (390px)** | Collection Match Mobile - Light | Collection Match Mobile - Dark |
| **Desktop (1280px)** | Collection Match Desktop - Light | Collection Match Desktop - Dark |

---

### 3. Multiple Candidates Selection Dialog

Dialog listing matching candidate spools using `CollectionSpoolCard`.

| Viewport / Theme | Light Theme | Dark Theme |
|---|---|---|
| **Mobile (390px)** | Candidate List Mobile - Light | Candidate List Mobile - Dark |
| **Desktop (1280px)** | Candidate List Desktop - Light | Candidate List Desktop - Dark |

---

## Verification

- **TypeScript Type Checks**: `npm run check` (PostgreSQL) and `npm run check:sqlite` (SQLite) pass with 0 errors.
- **ESLint**: `npm run lint` passes with 0 errors and 0 warnings (`--max-warnings 0`).
- **Unit & Component Tests**: All 60 Vitest tests pass across `collection-spool-card.test.tsx`, `collection-scanner-translations.test.ts`, and `filament-modal.test.tsx`.
- **Playwright E2E Tests**:
  - `npm run test:e2e tests/e2e/priority-0-mobile-decision.spec.ts`: 3/3 passed (mobile scanner decision workflow, catalog fallback, conflict candidates).
  - `npm run test:e2e tests/e2e/toast-auto-dismiss.spec.ts`: 3/3 passed (toast lifecycle, mobile touch auto-dismiss, viewport bounds and touch close button).


Azumi0:
### Update: Toast Auto-Dismiss Lifecycle & Mobile Touchscreen Fix

During testing, an issue was identified where toast notifications ceased auto-dismissing automatically and had to be closed manually.

#### Root Cause Analysis (Systematic Debugging)
1. **Radix UI `isClosePausedRef` Trap**:
   - `@radix-ui/react-toast` maintains a single shared `isClosePausedRef` on the root `ToastProvider`.
   - Any user interaction (such as tapping on a mobile touchscreen, clicking the `<ToastClose />` button, swiping, or window blur) sets `isClosePausedRef.current = true` to pause dismissal.
   - Because mobile touchscreens do not fire `pointerleave` events (and elements unmounting on dismissal do not emit `pointerleave`), `isClosePausedRef.current` remained stuck at `true` indefinitely for the remainder of the session.
   - As a result, all subsequent toasts checking `if (open && !context.isClosePausedRef.current) startTimer(duration)` completely skipped starting their auto-dismiss timer, causing them to stay on screen until manually closed.
2. **Missing Timer Lifecycle in `use-toast.ts`**:
   - `use-toast.ts` previously only managed an unmount delay (`TOAST_REMOVE_DELAY`) *after* a toast had already been dismissed. It had no auto-dismiss timer of its own and delegated 100% of dismiss scheduling to Radix's DOM event listeners.

#### Solution
- **Guaranteed Auto-Dismiss Timer**: Added an explicit auto-dismiss timer lifecycle (`scheduleAutoDismiss`) directly in `client/src/hooks/use-toast.ts` with a default 5000ms timer (`DEFAULT_TOAST_DURATION = 5000`).
- **Configurable & Clean Cancellation**: Supports custom duration overrides (e.g. `duration: 3000` or `duration: Infinity`), immediately cancels pending timers upon manual closure (`<ToastClose />`), and renews the timer if a toast is updated via `toast().update(...)`.
- **Explicit Fallback**: Added `duration={5000}` to `<ToastProvider>` in `client/src/components/ui/toaster.tsx` as defense in depth.
- **Code Review Polish**: Added `break-words` on subtitle text in `CollectionSpoolCard` and standardized candidate `DialogFooter` button ordering to `sm:flex-row-reverse` per `AGENTS.md` guidelines.

#### Verification
- Added unit tests in `tests/lib/use-toast.test.ts` covering default 5s auto-dismiss, custom durations, `Infinity` persistence, immediate manual dismiss cancellation, and update renewal.
- Added Playwright E2E tests in `tests/e2e/toast-auto-dismiss.spec.ts` covering 5s auto-dismiss, manual close button clicks, subsequent toast dismissal, and mobile touchscreen interaction.
- All 56 Vitest suites (684 tests) and all Playwright specs (`priority-0-mobile-decision` and `toast-auto-dismiss`) pass cleanly.

Azumi0:
### Update: Fix Datepicker Validation Failure for `lastDryingDate` and `purchaseDate`

During testing on desktop and mobile viewports, an issue was discovered where selecting a date from the datepicker on "Add Filament" or "Edit Filament" modals caused form submission to fail with:
> `Validation error: must be a date in YYYY-MM-DD form at "lastDryingDate"` (or at `"purchaseDate"`)

#### Root Cause Analysis
1. **Payload Serialization Mismatch**:
   - The datepicker (`Calendar` / `react-day-picker`) stores a JavaScript `Date` instance in React Hook Form.
   - On submission, `onSubmit` spread the raw `data` into `payload`.
   - When serialized with `JSON.stringify(payload)` during `apiRequest`, `Date.prototype.toJSON()` converted the dates into full ISO-8601 strings (e.g. `"2026-09-17T00:00:00.000Z"`).
   - The backend route validates against `filamentWriteSchema` / `filamentPatchSchema`, which strictly enforced `isoDate = z.string().regex(/^\d{4}-\d{2}-\d{2}$/, "must be a date in YYYY-MM-DD form")`.
   - The presence of the `"T..."` timestamp component caused Zod to throw validation errors and reject the request with HTTP 400.
2. **Timezone Offset Shift**:
   - `new Date("YYYY-MM-DD")` in JavaScript creates a UTC midnight date. In timezones west of UTC (e.g. US Eastern/Pacific), formatting this locally caused the date to shift backward by 1 day (e.g. September 17 displayed as September 16).
3. **Clearing Dates on Existing Spools**:
   - Deselecting a date previously set the field to `undefined`, which `JSON.stringify` drops. In `PATCH /api/filaments/:id`, the backend ignored omitted fields, preventing users from clearing existing dates in the database.

#### Solution
- **Client-Side Formatting (`formatDatePayload`)**:
  - Automatically formats `Date` instances to `"YYYY-MM-DD"` in local time before submitting.
  - Distinguishes between `null` (emitted on `isEditing = true` when clearing a date so PATCH sets the DB column to `NULL`) and `undefined` (on new spools so POST defaults to NULL).
  - Normalizes any existing strings by splitting on `"T"` prior to parsing, preventing negative UTC offset day-shift bugs.
- **Timezone-Safe Date Parsing (`parseDateValue`)**:
  - Uses `date-fns/parseISO` on the date component to initialize dates at local midnight across all timezones without timezone-drift.
- **Datepicker UI & Touch Targets**:
  - Added an accessible clear button (`t("common.none")`) in the calendar popovers meeting mobile accessibility guidelines (`min-w-[44px]`, `h-9` + generous padding).
  - Wrapped with `<PopoverClose asChild>` so the popover closes cleanly when cleared.
- **Backend Defense-in-Depth (`shared/schema.ts`)**:
  - Enhanced `isoDate` schema to validate standard `"YYYY-MM-DD"` strings as well as full ISO-8601 timestamps (normalizing them to `"YYYY-MM-DD"`), while strictly rejecting invalid values like `"yesterday"`.

#### Verification
- **Integration Tests** (`tests/routes/filament-writes.test.ts`): Verified that POST accepts both `"YYYY-MM-DD"` and ISO timestamps for `lastDryingDate` and `purchaseDate`, and PATCH clears them when `null` is sent.
- **Component Tests** (`tests/components/filament-modal.test.tsx`): Added unit tests for `parseDateValue` and `formatDatePayload`, verifying local date parsing and formatting across timezones (including execution under `TZ="America/New_York"`).
- **Full Suite**: All 56 Vitest test suites (694 tests) pass cleanly.
- **Type Check & Lint**: `npm run type-check` and `npm run lint` pass with 0 errors.

the-luap (changes requested):
Thanks for the thorough work here — the mobile problems this fixes are real, and two of the root causes you found are ones I'd have missed: the duplicate `<Toaster />` in `main.tsx` (every toast was rendered twice) and the stuck Radix `isClosePausedRef`. `CollectionSpoolCard` replacing three copies of the same card markup is a clear improvement.

I checked the timezone risk in the date fix carefully, since that's where this class of bug usually hides. `fff1313` splits on `T` before parsing, so it holds up: I ran the suites under `TZ=America/New_York` and `TZ=Pacific/Kiritimati` and both pass. Type check, lint, the i18n suites and the component suites are clean locally, and CI is green.

My main request is about scope rather than any single line.

## Please split this PR

It is now +1525/-295 across 25 files and contains five separable changes: the mobile redesign, the toast lifecycle fix, a change to how every API error is thrown, the datepicker fix, and repo policy docs. Each has a different risk profile and a different reason to revert. The datepicker fix in particular is self-contained, well tested and could merge on its own today.

## Blocking

**1. The API error change is app-wide and hidden in a UI PR** — `client/src/lib/queryClient.ts`

Before, `throw jsonData` sat inside a `try` whose `catch` swallowed it, so every failed request surfaced as `Error("404: {json}")`. Now a JSON error body is thrown as a plain object whose `message` is the server's text. That changes the error text in every toast and `onError` in the app, not just in the scanner.

One consequence visible today: `apiRequest`'s "don't log 401s on auth endpoints" check is `error instanceof Error && error.message.includes('401')`. A plain object fails both halves, so a failed login now writes to `console.error`.

I am not saying the change is wrong — it's arguably a fix for an old bug. But it deserves its own PR with a test that pins the contract, because the blast radius is every call site.

**2. Guessing "not found" from the error text** — `filament-modal.tsx:1232`

```ts
const isNotFound = err?.status === 404 || (typeof err?.message === "string" && err.message.includes("404"));
```

Since change 1 sets `err.status` on every thrown error, the string check is redundant. It also misfires: any non-JSON error body containing "404" — a barcode, a proxy error page — is silently treated as "not found", so the user gets "not found in any source" instead of an error. `err?.status === 404` alone is enough.

## Non-blocking

**3. `AGENTS.md`, the new `CLAUDE.md` section, and `docs/superpowers/plans/…`.** I'd rather these were a separate PR we can discuss on their own. As written, `AGENTS.md` pins exact Tailwind class strings and names "PR #40" as the canonical reference; both go stale the first time a component is restyled, and a stale standard is worse than none. If we keep a mobile standard, I'd write it as intent ("no auto-dismissing UI for decisions", "no hover-only controls", "test down to 320px") and let the code be the reference.

**4. The toast close button is hidden on tablets** — `toast.tsx:80`

`opacity-80 sm:opacity-0` means any touch device at 640px or wider (an iPad, most Android tablets) gets the invisible-until-hover close button again, which is the pattern `AGENTS.md` §4 forbids. Gate on hover capability instead of width:

```
opacity-80 [@media(hover:hover)]:opacity-0 [@media(hover:hover)]:group-hover:opacity-100
```

**5. The auto-dismiss timer does not pause on hover or focus.** Radix pauses its timer while a toast is hovered or focused; the new `scheduleAutoDismiss` timer does not, so a toast can disappear while it is being read or while a keyboard user has it focused. No toast uses an action button anymore, so nothing is unreachable, but long error messages can vanish mid-read. Pausing on the viewport's `pointerenter`/`focusin` — or leaving `duration` to Radix now that the duplicate provider is gone — would fix it.

**6. `tests/lib/use-toast.test.ts` cleanup doesn't clean up.** `toast({ id: t.id } as any).dismiss()` creates a *new* toast (`toast()` overwrites the id it is given) and dismisses that one, leaving the lingering toast in `memoryState`. It passes today because `TOAST_LIMIT` keeps the list short, but it isn't doing what it looks like.

**7. `isoDate` is now exported from `shared/schema.ts` but nothing imports it.** It can stay module-private.

Happy to re-review quickly once the split lands — the scanner redesign and the date fix both look mergeable on their own.

Azumi0:
Split this PR into focused, independent PRs per the review feedback, and addressed Note 2 directly on this branch:

### Separated PRs:
1. **Datepicker timezone parsing & validation fix** -> https://github.com/the-luap/filadex/pull/45 (#45)
   - Self-contained fix for `lastDryingDate` and `purchaseDate` parsing by splitting on `T` before `parseISO`, `PopoverClose` for calendar reset, and write API tests.
   - Per **Note 7**, `isoDate` remains module-private in `shared/schema.ts`.

2. **API error contract & silent 401 handling** -> https://github.com/the-luap/filadex/pull/46 (#46)
   - Per **Note 1**, introduces `ApiError extends Error` with `status` and `data` properties, preserving `instanceof Error` across all call sites and ensuring 401s on `/api/auth/*` routes are not logged to `console.error`.
   - Adds contract test suite in `tests/lib/queryClient.test.ts`.

3. **Toast lifecycle & tablet accessibility** -> https://github.com/the-luap/filadex/pull/47 (#47)
   - Resolves duplicate `<Toaster />` in `main.tsx`.
   - Per **Note 4**, uses `[@media(hover:hover)]` for tablet close button visibility.
   - Per **Note 5**, implements hover/focus pause & resume timer lifecycle wired to `<ToastViewport>`.
   - Per **Note 6**, adds clean memory reset and dismiss helpers in `use-toast.ts` and updates tests.

4. **Intent-based mobile guidelines** -> https://github.com/the-luap/filadex/pull/48 (#48)
   - Per **Note 3**, adds intent-based guidelines to `AGENTS.md` and `CLAUDE.md` without rigid class names or ephemeral PR references, and removed scratch planning documents.

---

### Changes Remaining in this PR (#44):
- Diff reduced from **+1525 / -295 across 25 files** down to **+751 / -187 across 11 files**.
- Strictly focused on `CollectionSpoolCard`, candidate selection cards, mobile-friendly dialogs, and replacing the transient Priority 0 toast with in-context decision dialogs.
- Per **Note 2**, `queryCommunityCatalogGtin` now checks `err?.status === 404` directly without guessing from error message strings.

the-luap (changes requested):
Thanks for splitting this — #45, #46, #47 and #48 are each reviewable on their own, and I've approved all four. What's left here is the scanner redesign, which is the right shape for this PR.

One blocking problem, and it's the cause of the red e2e run.

## The 404 check now always fails, so "not found" is reported as an error

`3e215e1` narrowed the check to `err?.status === 404` (my Note 2). That's the right destination, but `status` only exists once #46 lands — on this branch `throwIfResNotOk` still throws `new Error("404: {json}")` with no `status` property. So `isNotFound` is always false, and a barcode that simply isn't in the catalog takes the error path.

Reproduced locally on `pr/44` head (`build:e2e`, then the failing spec):

```
TOAST after 53ms: "Error"
NET: 404 /api/community-filaments/gtin/9999999999999?diameter=1.75&weightGrams=1000
CONSOLE: API request to ... failed: Error: 404: {"message" ...
```

The user scans an unknown barcode and gets a red "Error / failed to search the community catalog" toast instead of "Barcode … not recognized (searched Filadex QR, Bambu Lab, and Open Filament Database)". The `not_found` fallback — including the collection-vs-catalog dialog paths that depend on it — never runs.

That is also precisely why `priority-0-mobile-decision.spec.ts:259` fails: it waits for the "not recognized" toast, which this branch can no longer produce.

**Fix:** merge #46 first, then rebase this branch on it. The check is already correct for that world; it just needs the branch to contain `ApiError`. (I'd rather not see the string sniff come back as a stopgap.)

## One test is in the wrong PR

The same spec's `mobile: toast renders within mobile viewport without horizontal overflow and close button is touch accessible` asserts toast *presentation* — overflow bounds and a tappable close button — which is #47's code, not this PR's. Even after the rebase it would be testing another PR's behaviour from here. Please move that test into #47 and keep this spec focused on the scanner dialogs and candidate cards.

## The rest looks good

`CollectionSpoolCard` replacing three near-identical card blocks is a clear win, and the static (`interactive={false}`) preview inside the decision dialogs is a nice touch — the same component that lists candidates also shows the matched spool. Replacing the auto-dismissing toast with an `AlertDialog` is the right call for a decision the user has to make.

I also checked that the widened base classes in `dialog.tsx` / `alert-dialog.tsx` don't disturb the larger modals: `filament-modal`, `settings-dialog` and `user-management-modal` all set their own width and `overflow-hidden`, and tailwind-merge keeps those winning.

Unit tests, `tsc` and `eslint` are clean on this branch (164 tests); the only failure is the e2e above.

the-luap:
#45, #46, #47 and #48 are all merged to `main` now, so this branch can be updated.

I verified the update actually resolves the blocker: I merged current `main` into this branch locally, ran `npm run build:e2e`, and the whole `priority-0-mobile-decision.spec.ts` file passes — **4/4, including the `toast renders within mobile viewport … close button is touch accessible` test that fails here today**. With `ApiError` from #46 in the branch, `err?.status === 404` resolves correctly, so an unknown barcode produces the "not recognized" toast again instead of the red error toast.

Git reports no conflicts between this branch and `main`, so `git merge origin/main` (or a rebase, whichever you prefer) should be all it needs.

Two things still open from my review:

1. **Move the toast test into its own home.** `mobile: toast renders within mobile viewport without horizontal overflow and close button is touch accessible` asserts toast presentation, which is #47's code and is now on `main`. It passes here only because the branch picks up those changes. It belongs in a toast spec rather than the scanner spec — the scanner spec should fail when the *scanner* breaks.
2. Once it's updated and CI is green, ping me and I'll re-review. Everything else in this PR looked good on my last pass.

the-luap:
One more thing while this branch is being updated: please add screenshots to the PR description.

I'm tightening `CONTRIBUTING.md` so that screenshots are required for any change a user can see, rather than "whenever possible". The reason is this PR's own history — a dialog overflowing a phone screen and a close button invisible without a mouse were both plainly visible in a screenshot and invisible in the diff. I can follow the class changes in `dialog.tsx` and `collection-spool-card.tsx` and still not know what the result looks like.

For this PR that means:

* The new collection-match decision dialog and the candidate-list dialog, at a phone width (390px) and at desktop.
* The before/after of a candidate card, since `CollectionSpoolCard` replaces three different card layouts — the swatch, badges and truncation behaviour are the substance of the change.
* Both themes, since the card adds explicit light/dark colours.

`npx playwright test --headed` against a seeded database gets you to those dialogs quickly; your `priority-0-mobile-decision.spec.ts` already walks through every one of them.

Azumi0:
Updated the branch and addressed all review points:

1. **Merged `main`**: Picked up #45, #46, #47, #48, and #49. With `ApiError` landed, `err?.status === 404` in `queryCommunityCatalogGtin` resolves as intended without false error toasts.
2. **Relocated Toast Test**: Moved `mobile: toast renders within mobile viewport without horizontal overflow and close button is touch accessible` into `tests/e2e/toast-auto-dismiss.spec.ts`. `tests/e2e/priority-0-mobile-decision.spec.ts` is now strictly scoped to the scanner modals and candidate cards.
3. **Desktop Dialog Polish**: Widened the decision dialogs to `max-w-xl` with `sm:flex-wrap` on `AlertDialogFooter` so multi-action button labels ("Search Community Catalog instead", "Use Collection Specs") do not truncate on desktop screens.
4. **UI Screenshots Added**: Generated and embedded all requested screenshots into the PR description:
   - Candidate card: Before vs. After comparison (swatch border, secondary material badge, outline pills, truncation/wrapping, 44px min touch target) in light and dark themes.
   - Collection-match decision dialog: Mobile (390px) and Desktop (1280px) in light and dark themes.
   - Multiple candidate spools dialog: Mobile (390px) and Desktop (1280px) in light and dark themes.

All 11 CI checks are passing. Ready for re-review @the-luap.

the-luap (approved):
Approving. Everything I asked for is here, and I verified the parts that mattered rather than taking the description's word for it.

**The blocker is gone.** With `main` merged in, `ApiError` is present and `err?.status === 404` resolves correctly. I rebuilt the branch locally (`npm run build:e2e`) and ran both specs: 6/6 pass, including the scanner suite that was red before. Unit suites pass (189 tests), `tsc` and `eslint` clean.

**The toast test found its home.** `priority-0-mobile-decision.spec.ts` is now strictly the scanner dialogs and candidate cards, and the presentation test lives in `toast-auto-dismiss.spec.ts` next to the code it exercises. That's the point — the scanner spec should now fail only when the scanner breaks.

**The screenshots earned their place.** All ten load. The desktop capture shows exactly why `177b115` was needed — "Search Community Catalog instead" sits on one line at `max-w-xl` — and the card comparison makes the swatch, badge and wrapping changes legible in a way the class diff never could. This is the first PR under the new `CONTRIBUTING.md` rule and it sets a good example for the next one.

Three notes, all non-blocking, none worth another round:

1. **Where the images live.** They're served from a `media` branch on your fork. If that branch or fork ever goes away, this PR's history loses its screenshots. Dragging images straight into the description box uploads them to GitHub's own attachment storage, which is tied to the PR rather than to a repo you might clean up later.
2. **The card "before" is a reconstruction,** not a capture of the old UI — a comparison page rendering both variants. It communicates the change well, so I'm not asking for a redo; just noting that a genuine before/after is more convincing when the claim is "this fixes a layout bug".
3. **`sm:flex-wrap` with the footer's inherited `sm:space-x-2`.** On desktop the wrapped "Cancel" now lands alone on a second row, and `space-x-2` applies its margins asymmetrically under `flex-row-reverse`. Purely cosmetic and visible in your own screenshot; worth a look next time you're in that file.

Merging.

PR #45, opened 2026-09-18
fix(filament-modal): resolve datepicker validation and timezone parsing for lastDryingDate and purchaseDate

Azumi0:
Extracted from #44 per code review request.

- Resolves datepicker form validation errors for `lastDryingDate` and `purchaseDate`.
- Safely parses dates by splitting on `T` before `parseISO` to avoid timezone shifts across daylight savings / timezones.
- Adds `PopoverClose` export to close calendar popovers when selecting "None".
- Keeps `isoDate` unexported / module-private in `shared/schema.ts` per review note 7.
- Adds unit and route tests covering timezone parsing, format payload, and write API validation.


the-luap (approved):
Approving. Clean, self-contained bug fix with the tests I'd want.

I paid particular attention to the timezone handling, since that's where this class of fix usually goes wrong. Splitting on `T` before `parseISO` is what makes it safe: `parseISO("2026-09-17")` builds local midnight, so `format(..., "yyyy-MM-dd")` returns the same day everywhere. I ran the suites under `TZ=America/New_York` and `TZ=Pacific/Kiritimati` (UTC-4 and UTC+14) and they pass in both, which is the real check — the tests would be green in UTC either way.

The `null` vs `undefined` distinction in `formatDatePayload` is the subtle part and it's right: `null` on edit so PATCH nulls the column, `undefined` on create so `JSON.stringify` drops the key. The route tests cover both directions, and the `PopoverClose` wrapper on the clear button means the popover doesn't sit open over a field the user just cleared.

Three notes, none blocking:

1. **`isoDate` is exported but nothing imports it.** It can stay module-private.
2. **Widening the server contract is now optional.** The client always sends `YYYY-MM-DD` after this fix, so `isoDate` accepting full ISO timestamps is defence-in-depth for a case that no longer occurs — and it's a contract you can't easily narrow later. Fine to keep deliberately; just worth being a deliberate choice.
3. **`lastDryingDate` gained `disabled={(date) => date > new Date()}`.** Sensible (you can't have dried a spool tomorrow), but it's a behaviour change beyond the reported bug and isn't mentioned in the description. Worth a line in the PR body so it isn't a surprise later.

Locally: 168 tests pass, `tsc` and `eslint` clean, CI green.

PR #46, opened 2026-09-18
fix(api): introduce ApiError with status and contract tests, fixing auth 401 logging

Azumi0:
Extracted from #44 per review note 1.

- Introduces `ApiError extends Error` in `client/src/lib/queryClient.ts` with `status` and `data` properties, preserving the `instanceof Error` prototype contract.
- Extracts server error message or detail while keeping `status` available for callers (e.g. `err?.status === 404`).
- Updates `apiRequest` and `getQueryFn` to correctly suppress `console.error` on 401s for `/api/auth/*` endpoints without misfiring on plain error objects.
- Adds comprehensive contract test suite in `tests/lib/queryClient.test.ts` pinning 200, 204, 401, 404, and 500 error behavior.


the-luap (approved):
Approving. This is the change I asked to be split out, and it came back better than what it replaced.

`ApiError` gives every failed request a real `status` instead of leaving callers to sniff the message text, the `Object.setPrototypeOf` call keeps `instanceof` working through the TS down-level emit, and the old structure — a `throw` inside a `try` whose own `catch` swallowed it — is gone rather than papered over. The 401 logging check now keys off `status`, with the string branch kept for the `Error('Not authenticated')` path that `apiRequest` still throws itself. That is the right call, not an oversight.

The test file is the part that makes this mergeable: it pins the contract (JSON `message`, JSON `detail`, non-JSON body, 204, and the logging behaviour on auth vs non-auth endpoints) rather than the implementation.

Verified locally: `tests/lib`, `tests/components` and `tests/i18n` pass (169 tests), `tsc` and `eslint` clean, and CI is green.

Two notes, neither blocking:

- `message` comes from the response body, so a server that answers `{"message": {...}}` would produce `"[object Object]"`. `String(message)` keeps it from throwing, so this is a cosmetic edge at worst.
- Please merge this **before** #44. I've left the details there: #44's narrowed 404 check depends on `ApiError.status` existing, and its e2e test fails without this PR.

PR #47, opened 2026-09-18
fix(toast): resolve duplicate provider, hover/focus pause lifecycle, and tablet close button

Azumi0:
Extracted from #44 per code review notes 4, 5, and 6.

- Removes duplicate `<Toaster />` provider in `client/src/main.tsx` (previously rendering every toast twice).
- Gates toast close button on hover capability via `[@media(hover:hover)]` in `toast.tsx` rather than screen width, making close buttons visible on tablets (note 4).
- Implements auto-dismiss timer pause and resume on hover/focus via `pauseAutoDismiss` and `resumeAutoDismiss` wired to `<ToastViewport>`, preventing toasts from vanishing mid-read (note 5).
- Exports clean `clearToasts()` and `dismiss(id)` helpers in `use-toast.ts` and cleans up `tests/lib/use-toast.test.ts` without creating dummy toasts (note 6).
- Adds unit tests for hover/focus pause/resume and e2e lifecycle tests in Playwright.


the-luap (approved):
Approving. The pause/resume lifecycle is a real improvement over what I saw in #44, and the root-cause work behind it is sound.

What convinced me:

- **The design avoids the trap it was written to fix.** Radix's bug was a single shared `isClosePausedRef` on the provider: once stuck true, *every future* toast skipped its timer. Here the pause state lives per timer in `toastTimers`, and `scheduleAutoDismiss` always starts a new toast unpaused. So even if a `pointerleave` is missed, the damage is limited to the toasts open at that moment, and the user can still close them by hand. That's the important structural difference.
- **The remaining-time arithmetic is correct across repeated pause/resume cycles** — `remaining` decrements by elapsed and `startTime` is reset on resume, rather than restarting the full duration each time.
- **`[@media(hover:hover)]` on the close button** fixes the tablet case properly: gating on hover capability rather than on viewport width.
- **`clearToasts()`** gives the tests a real reset instead of the id trick that didn't actually dismiss anything.

I checked the touch path that motivated the change: in Chromium with touch emulation, tapping an element fires `pointerenter → pointerdown → pointerup → pointerleave`, so a tap pauses and then resumes rather than leaving the timer stuck. The pause/resume handlers also sit on the viewport, which stays mounted across toasts, so an unmounting toast can't strand the pause either.

Two suggestions, neither blocking:

1. **There are now two timers per toast.** `<ToastProvider duration={5000}>` keeps Radix's timer running alongside `scheduleAutoDismiss`, so both race to close the same toast. Harmless today (both are 5s and both pause on hover), but it means a future change to `DEFAULT_TOAST_DURATION` silently won't take effect until it exceeds 5000ms, and per-toast `duration: Infinity` still gets a 5s Radix timer. Passing `duration={Infinity}` to the provider would make `use-toast` the single owner of dismissal.
2. **`TOAST_REMOVE_DELAY` drops 5000 → 1000.** Correct — it only governs removal from memory after the close animation — but it's an unrelated tuning change worth a line in the description.

Locally: 168 tests pass, `tsc` and `eslint` clean, CI green.

PR #48, opened 2026-09-18
docs: add intent-based mobile-first UI guidelines in AGENTS.md and CLAUDE.md

Azumi0:
Extracted from #44 per review note 3.

- Adds `AGENTS.md` documenting intent-based mobile-first design principles:
  1. Testing down to 320px width without horizontal overflow.
  2. No auto-dismissing notifications for interactive decisions.
  3. Touch targets >= 44x44px and no hover-only controls.
  4. Vertically stacked dialog actions for thumb reach on mobile.
  5. Text container safety and overflow prevention.
- Points to `AGENTS.md` from `CLAUDE.md` §6.
- Keeps standards focused on intent rather than brittle CSS class strings or ephemeral PR numbers.


the-luap (approved):
Approving. This is exactly the rewrite I hoped for: the rules now state intent ("no auto-dismissing UI for decisions", "no hover-only controls", "verify down to 320px") instead of pinning class strings and a PR number that would go stale the first time a component is restyled. §3's `@media(hover:hover)` escape hatch is the right level of detail — it names the technique without prescribing the exact utility.

One small thing, non-blocking: `AGENTS.md` and the new `CLAUDE.md` section now state the same rules twice, so they can drift apart. Since `CLAUDE.md` §6 already points at `AGENTS.md`, I'd trim it to just the pointer and let `AGENTS.md` be the single source.

PR #50, opened 2026-09-18
fix(ui): use flex gap in dialog footers and establish PR asset workflow

Azumi0:
## Description

Follow-up to pull request #44 addressing the three non-blocking review notes left by @the-luap:

1. **Native GitHub Attachment Hosting (Note 1):**
   - Implemented `scripts/upload-pr-asset.sh` and npm script `npm run upload:pr-asset` to upload UI screenshots and videos directly to GitHub's native `user-attachments` backend (`https://uploads.github.com/user-attachments/assets`), outputting canonical URLs and Markdown embed tags.
   - Replaces ephemeral `media/*` or fork branch hosting so pull request histories never lose their visual evidence if a branch or fork is deleted or cleaned up.
   - Includes URL-encoding for filenames with spaces/special characters, case-insensitive MIME detection, support for video formats (`mp4`, `webm`, `mov`), and robust multi-file error handling with non-zero exit propagation.
   - Added direct rules to `AGENTS.md`, `CLAUDE.md`, and `CONTRIBUTING.md` instructing AI agents and contributors to always use native attachments.

2. **Genuine Before/After Baseline Captures (Note 2):**
   - Established strict rules across `AGENTS.md`, `CLAUDE.md`, and `CONTRIBUTING.md` requiring genuine baseline "before" captures from the unmodified base branch (`main`) before making any code edits (or via a separate git worktree of `main`).
   - Forbids synthetic comparison pages or side-by-side mockup components that emulate the old UI on the feature branch.
   - Added `.screenshots/` to `.gitignore` so local capture scratch directories remain uncommitted.

3. **Dialog & Sheet Footer Spacing Fix (Note 3):**
   - Replaced margin-based sibling spacing `sm:space-x-2` with flex gap `sm:gap-2` in `AlertDialogFooter` (`alert-dialog.tsx`), `DialogFooter` (`dialog.tsx`), and `SheetFooter` (`sheet.tsx`).
   - Eliminates the Tailwind margin bug where `space-x-2` injected asymmetric margins under `flex-row-reverse` and leaked margins onto wrapped items (causing buttons like "Cancel" to sit alone with an extra left margin on wrapped rows).
   - Documented the `gap-*` rule in `AGENTS.md` and `CLAUDE.md`.

Ref: #44

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- **TypeScript Type Checks:**
  - `npm run check` (PostgreSQL): 0 errors
  - `npm run check:sqlite` (SQLite): 0 errors
- **Linter:**
  - `npm run lint` (ESLint): 0 errors, 0 warnings (`--max-warnings 0`)
- **Unit & Integration Tests:**
  - `npm test` (Vitest): All 57 test suites passed (705 passed, 9 skipped, 0 failed).
- **PR Asset Upload Script:**
  - Tested uploads against GitHub's native user-attachments API with single and multiple files.
  - Verified URL encoding with filenames containing spaces and ampersands (`Screenshot 2026-09-18 at 14.41.00 & test.png`).
  - Verified uppercase extension handling (`.PNG`) and error propagation (exits with code 1 when files are missing or upload fails).

## Checklist:


the-luap (changes requested):
Thanks for taking all three notes seriously — especially the hosting one, which I raised as a "next time" and you turned into something the next contributor can actually run.

Good news first, because it's the part I was least sure of: **the upload endpoint really does work with a `gh auth token`.** I probed it directly rather than assuming — a deliberately bad token returns `401 Bad credentials`, and a valid one returns `422 Validation Failed` on an empty body. So the token is accepted and the request routes; the approach is sound.

One blocking problem.

## The script cannot run on macOS

`${ext,,}` (line 90) is a bash 4 lowercase expansion. macOS ships bash **3.2.57** as `/bin/bash`, and `#!/usr/bin/env bash` finds exactly that unless someone has installed a newer bash by hand. Running it on this machine:

```
$ ./scripts/upload-pr-asset.sh /tmp/tiny.png
Uploading using repository context: the-luap/filadex (ID: 981343391)
./scripts/upload-pr-asset.sh: line 90: ${ext,,}: bad substitution
```

It aborts for every file, on every invocation, on the platform I review from. `npm run upload:pr-asset` runs it through `bash` explicitly, so that path fails the same way.

The fix is one line:

```sh
case "$(printf '%s' "$ext" | tr '[:upper:]' '[:lower:]')" in
```

Worth a quick grep for other bashisms while you're in there — I didn't spot any others (`+=` on arrays and `[[ =~ ]]` are both fine in 3.2), but I only read for it, I didn't test each branch.

## Smaller things

1. **No size check before upload.** GitHub caps user attachments at 10MB and answers with an HTML-laced validation message — I got `"size Yowza that's a big file. <span class='drag-and-drop-error-info'>…"` in my probe. Since this is meant for screen recordings, a `stat`-based check with a plain message would beat surfacing that blob to the user.
2. **`AGENTS.md` has a dangling `# or`** at the end of the upload snippet with nothing after it — presumably an alternative invocation got trimmed.
3. **The repo-candidate loop could be one call.** `GITHUB_REPOSITORY` → `origin` → `upstream` → a hardcoded `the-luap/filadex` is a lot of fallback for something `gh repo view --json id -q .id` answers directly from the current checkout. The hardcoded fallback in particular will be wrong the moment someone forks this.

## The rest

The `space-x-2` → `sm:gap-2` change in `AlertDialogFooter`, `DialogFooter` and `SheetFooter` is the right fix and correctly scoped — the `sm:` prefix means the stacked mobile layout is untouched, and callers that already pass their own `gap-2` are unaffected. Unit tests (157), `tsc` and `eslint` are clean on the branch.

The documentation changes land well. Trimming `CLAUDE.md` §6 to a pointer at `AGENTS.md` is exactly what I asked for in #48, and "never serve PR screenshots from a git branch" states the reason rather than just the rule.

Azumi0:
Addressed review feedback in b61e545:

1. **macOS Bash 3.2 Compatibility:**
   - Replaced Bash 4 lowercase parameter expansion `${ext,,}` in `get_mime()` with `case "$(printf '%s' "$ext" | tr '[:upper:]' '[:lower:]')" in`.
   - Verified remaining script syntax (`local`, array appends `+=`, regex matching `[[ =~ ]]`) is fully supported under Bash 3.2.

2. **Pre-Upload File Size Check:**
   - Added `get_file_size()` supporting macOS (`stat -f %z`) and Linux (`stat -c %s`) with `wc -c` fallback.
   - Files exceeding GitHub's 10MB attachment ceiling (10,485,760 bytes) are caught locally with a clear error message (`Error: File exceeds GitHub 10MB attachment limit: ...`) and increment `FAILED_COUNT` without triggering GitHub's HTML validation response.

3. **AGENTS.md Snippet:**
   - Removed the dangling `# or` and redundant bash invocation from the upload code block in [AGENTS.md](https://github.com/the-luap/filadex/blob/fix/pr-screenshot-workflow-and-footer-gap/AGENTS.md), keeping `npm run upload:pr-asset -- <files...>`.

4. **Repository Resolution & Upload Permissions:**
   - Removed the hardcoded fallback `the-luap/filadex` and the candidate array loop.
   - The script resolves candidate remotes prioritizing `origin` (the user's writable fork in contributor workflows, or the primary repo for maintainers), then `upstream`, `GITHUB_REPOSITORY`, and `gh repo view`.
   - Regarding `gh repo view --json id -q .id`: `gh repo view --json id` returns a GraphQL Node ID (e.g. `R_kg...`). The `uploads.github.com/user-attachments/assets?repository_id=...` endpoint rejects GraphQL IDs with `404 Not Found` and requires the numeric REST database ID. The script uses `gh api "repos/${ACTIVE_REPO}" --jq .id` to retrieve this integer ID.

the-luap (approved):
`b61e545` and `35d3fe2` clear the blocker. I re-ran the script from a macOS checkout rather than reading the diff, and this time it works end to end.

```
$ /bin/bash --version          # 3.2.57(1)-release
$ /bin/bash -n scripts/upload-pr-asset.sh          # parses clean
$ /bin/bash scripts/upload-pr-asset.sh "/tmp/PR 50 probe & test.PNG"
Uploading using repository context: the-luap/filadex (ID: 981343391)
Uploaded: /tmp/PR 50 probe & test.PNG
URL: https://github.com/user-attachments/assets/6c14654b-…
Markdown: ![PR 50 probe & test](https://github.com/user-attachments/assets/6c14654b-…)
```

That one filename exercises three of the things you fixed or claimed at once — a space, an ampersand, and an uppercase extension — and the `tr` replacement resolves the MIME type correctly. The error paths behave too: no arguments exits 1 with usage, a missing file reports and exits 1, and an 11MB file is caught locally with `Error: File exceeds GitHub 10MB attachment limit: … (11.00MB / 11534336 bytes)` instead of GitHub's HTML blob.

**You were right and I was wrong about `gh repo view --json id`.** It returns the GraphQL node ID (`R_kg…`), which that endpoint rejects; `gh api repos/{owner}/{repo} --jq .id` is the correct way to get the numeric database ID. Thanks for checking instead of just applying it — and for leaving the reason in a comment above the call, which is where the next person will need it.

The lazy resolution reads better too: `origin` first is right for the fork workflow, and dropping the hardcoded `the-luap/filadex` means a fork won't silently upload into someone else's repository context.

One observation, not a defect: an uploaded asset URL **404s to an anonymous fetch until it's referenced** in a PR body or comment. So a contributor who uploads and then checks the link in a logged-out tab may think it failed. Worth one line in `AGENTS.md` next to the upload snippet — "the URL resolves once the description is saved" — to save someone that confusion.

Unit tests (157), `tsc` and `eslint` clean on the branch; CI green. Approving.

PR #51, opened 2026-09-18
fix(toast): deduplicate timers by making use-toast single timing authority

Azumi0:
## Summary

Consolidates toast lifecycle auto-dismiss timers into `use-toast.ts` (`scheduleAutoDismiss`) as the single timing authority, disabling Radix UI's internal fallback timer and eliminating dual-timer race conditions.

Documented in detail in [ADR 0014: Toast Lifecycle Timer Deduplication and Single Source of Timing Authority](docs/adr/0014-toast-lifecycle-timer-deduplication.md).

## Problem

`<ToastProvider duration={5000}>` in `Toaster` initialized Radix's context duration to 5000ms. When toasts were rendered (`<Toast key={id} {...props}>`), Radix evaluated `const duration = durationProp || context.duration`.
- For default toasts (`durationProp === undefined`), Radix scheduled a 5000ms timer.
- For custom durations (e.g. 3000ms), Radix also scheduled a 3000ms timer.
- Meanwhile, `use-toast.ts` (`scheduleAutoDismiss`) ran its own concurrent timer handling pauses on hover/focus, updates, and delayed removal (`addToRemoveQueue`).

This caused two timers to race for every toast. A future increase to `DEFAULT_TOAST_DURATION` (e.g. to 8000ms) was silently intercepted by Radix's 5000ms timer closing the toast early, and persistent toasts (`duration: Infinity`) were vulnerable to fallback timers.

## Solution

1. **Disable Radix Fallback Timer**: Configured `<ToastProvider duration={Infinity}>` in `client/src/components/ui/toaster.tsx`. Radix's internal `startTimer` explicitly checks `if (!duration || duration === Infinity) return;`, bypassing internal timer creation.
2. **Omit Per-Toast Duration from Radix**: Extracted `duration: _duration` when spreading props to `<Toast key={id} {...props}>` so Radix never receives a numeric duration prop.
3. **Single Source of Truth**: All auto-dismiss timing, pause/resume calculations on hover/focus, and cleanup delays are managed solely by `use-toast.ts`.
4. **ADR 0014**: Authored `docs/adr/0014-toast-lifecycle-timer-deduplication.md` documenting this architectural decision.
5. **Testing**:
   - `tests/components/toaster.test.tsx`: Verifies `ToastProvider` receives `duration={Infinity}` and `<Toast>` elements receive `duration: undefined`.
   - `tests/lib/use-toast.test.ts`: Added test asserting that durations exceeding 5000ms (e.g. 8000ms) remain open past 5000ms and close at their designated time.

## Verification

- **TypeScript Checks**: `npm run check` (PostgreSQL) and `npm run check:sqlite` (SQLite) pass with 0 errors.
- **ESLint**: `npm run lint` passes with 0 errors and 0 warnings (`--max-warnings 0`).
- **Unit & Component Tests**: 58 test files passed (708 passing tests, 9 skipped).
- **Bundle Affordance Verification**: `npm run build && npm run verify:bundle` clean.
- **Playwright E2E**: `npm run build:e2e && npx playwright test tests/e2e/toast-auto-dismiss.spec.ts` passed (3/3).


the-luap (approved):
Approving. Small, precise, and it removes a real ambiguity rather than a hypothetical one.

I checked the claim the whole change rests on against the installed Radix source (`@radix-ui/react-toast@1.2.23`) rather than taking the ADR's word for it:

- `dist/index.mjs:314` — `const duration = durationProp || context.duration;`
- `dist/index.mjs:326` — `if (!duration2 || duration2 === Infinity) return;`

So `duration={Infinity}` on the provider plus omitting `duration` from each `<Toast>` does exactly what the ADR says: Radix schedules no `setTimeout` at all, and `use-toast` owns dismissal outright. The `8000ms` test is the right one to add — it fails against the old wiring, which is what makes it a regression test rather than a restatement.

ADR 0014 is the next free number, and the write-up explains the reasoning well enough that the next person won't "helpfully" restore `duration={5000}`.

Locally: `tests/lib/use-toast.test.ts` and `tests/components/toaster.test.tsx` pass (11 tests), `tsc` and `eslint` clean, CI green.

Two notes for the ADR rather than the code, neither blocking:

1. **The consequences section could name what was given up.** Radix's timer was also a backstop: if `use-toast`'s timer is ever paused and never resumed (a missed `pointerleave`), nothing else will close that toast now. I think that's the right trade — the pause is per timer, new toasts are unaffected, and the user can still close it by hand — but it's the kind of thing worth writing down next to the benefits.
2. **`tests/components/toaster.test.tsx` asserts the wiring, not the behaviour.** It mocks Radix and checks which props it receives, so it would still pass if Radix changed how it reads them. The `use-toast` duration tests are what actually pin the behaviour; the e2e toast spec covers the real DOM. Fine as a guardrail, just don't mistake it for coverage of the timer itself.

Azumi0:
Addressed the non-blocking notes in 0916f69:

1. **Secondary Backstop Trade-off Documented in ADR 0014:**
   - Added an explicit trade-off section to [ADR 0014](https://github.com/the-luap/filadex/blob/fix/toast-timer-deduplication/docs/adr/0014-toast-lifecycle-timer-deduplication.md) detailing the loss of Radix's secondary dismissal backstop (e.g. if a toast is paused on hover and misses a `pointerleave` event).
   - Documented the rationale for accepting this trade-off: pauses remain scoped per toast timer, new toasts are unaffected, manual dismissal remains available via close button and keyboard, and eliminating silent duration truncation outweighs relying on an accidental fallback timer.

2. **Test Scope & Boundary Clarification:**
   - Updated the Verification & Guardrails section of ADR 0014 to explicitly categorize `tests/components/toaster.test.tsx` as a structural JSX wiring guardrail.
   - Added explanatory commentary to `tests/components/toaster.test.tsx` stating that it guards prop wiring against regressions rather than verifying Radix's runtime consumption or timer mechanics, noting that behavioral verification is pinned by `tests/lib/use-toast.test.ts` and the Playwright e2e suite.

the-luap:
`0916f69` covers both notes, and the trade-off section is better than what I asked for — the three-point "accepted because" reads like something a future maintainer can actually argue with, rather than a disclaimer. Recording that Radix's timer *was* a backstop, and that losing it is deliberate, is exactly what keeps someone from "fixing" this by restoring `duration={5000}` in a year.

The verification section now separates the three layers properly — wiring guardrail, timing behaviour, real DOM — and the header comment in `toaster.test.tsx` says the same thing where somebody editing that file will see it.

Re-ran the suites on the new head: 11 tests pass, CI green. My approval stands — nothing further from me.

